PR #15, opened 2020-12-16
Add a github actions workflow for linting, testing and deploying workflows

bernt-matthias:
I hope that this is a good start. Not sure yet about deployment .. which is completely untested.

An example run can be seen here https://github.com/bernt-matthias/iwc/actions/runs/426163830


bernt-matthias:
shall be replaced by https://github.com/mvdbeek/workflow-testing-repo/

PR #17, opened 2021-02-21
Add pe-illumina-wgs workflow

mvdbeek:


bwlang:
Looks like the planemo test via github workflows is working nicely - yay - first workflow!
It's not so easy to evaluate the workflow in this interface though.
What's the best path here?  Manually import the .ga file to usegalaxy.org for evaluation?

workflows/sars-cov-2-variant-calling/pe-illumina-wgs/CHANGELOG.md
```diff
@@ -0,0 +1,3 @@
+0.1
+---------
+- Initial version of COVID-19: variation analysis on WGS SE data workflow
```
wm75, line 3: PE instead of SE?

workflows/sars-cov-2-variant-calling/pe-illumina-wgs/README.md
```diff
@@ -0,0 +1,5 @@
+COVID-19: variation analysis on WGS PE data
+-------------------------------------------
+
+This workflows performs paired end read mapping with bwa-mem followed by
+sensitive variant calling across a wide range of AFs with lofreq
```
wm75, line 5: ... and variant annotation with snpEff 4.5covid19.

mvdbeek:
@bwlang We can push some static artefacts, like the d3 visualization and maybe a screenshot, but for now that's probably the best option to look at the workflow.

mvdbeek:
Also the CVMFS apt repo seems to refuse connection frequently, maybe we need to download and install the .deb package manually :(.

bgruening:
> @bwlang We can push some static artefacts, like the d3 visualization and maybe a screenshot, but for now that's probably the best option to look at the workflow.

We should annotate the supporting Galaxy instances. This can happen for now manually (later automatically). But this would make inspection etc easier.

bwlang:
I think a screenshot is a good start, but to do a meaningful review I think I'd need to interact with the individual steps.
Maybe I can just get used to .ga format for those details. If we expect the ga to be easily importable to usegalaxy.org - maybe we could extend https://usegalaxy.org/workflows/import to allow import from a url and add a query param (e.g. ?import_url=https://raw.githubusercontent.com/galaxyproject/iwc/b93e1d0da084cf29f798ed918ffa80b261a08eb1/workflows/sars-cov-2-variant-calling/pe-illumina-wgs/pe-wgs-variation.ga)
to make it easy to import and review...

challenge: dependencies all available?

@mvdbeek Related to your concern about cvmfs failures, I also wonder about how reliable this will be with outside refs to input data (do we hit that too hard? is the URL durable?).  Should we try to make this work with a small enough input dataset that it can be easily added to the test artifacts like the genome is?

mvdbeek:
I think we can decide that on a case by case basis. The test data should be reasonably small, but here were downloading from the SRA's public GCP bucket, which should be fine. Small datasets on zenodo should als be fine, we've used that during the GTN smorgasbord training and I don't think there were any issues there.

mvdbeek:
Alright, the CVMFS apt server seems more stable today.

wm75 (approved):

PR #22, opened 2021-02-22
Initial version of ONT ARTIC workflow

mvdbeek:


bwlang:
I just tried this out on real dataset... I was surprised that there is no adapter trimming in this workflow.
Minimap will softclip, but there are still differences +/- porechop in the amplicon boundary regions.
Porechop is not awesome either... perhaps some better ONT adapter trimmer should be added?
image

bwlang:
@mvdbeek any idea why no adapter trimming is done for ONT?

wm75:
@bwlang as it says in the README, this WF performs, essentially, the same steps as that pipeline’s minion command, which does not include demultiplexing/adapter removal.
The ARTIC network recommends Guppy for upstream processing, which we currently don't have a Galaxy wrapper for. Porechop, as you're saying, would be one alternative, but whatever we include would mean we're deviating from the ARTIC network's [protocol](https://artic.network/ncov-2019/ncov2019-bioinformatics-sop.html).
My reasoning when assembling this WF was that users would need guppy anyway for base calling so they could just as well use it for read preprocessing, too.
If there are good solutions for base calling/demultiplexing/adapter removal, we could always bundle these in a separate ONT preprocessing WF.

bwlang:
Hi @wm75 
Is the goal to faithfully replicate part of the ARTIC protocol or to do the best possible analysis?  It sounds like we agree that it's a good practice to trim adapter sequences that are not expected in the reference before alignment.  While *some* instruments are configured to trim using guppy as suggested in the protocol, I think that's far from universal (e.g. these reads were acquired on a gridion with default minknow settings).

 I see a benefit to adding adapter trimming, though porechop alone does not look sufficient. Is it worth some energy to try to find better tools/settings to optimize trimming and integrate them here or do you think it's better to ask users to run multiple workflows?

wm75:
I would answer yes to most of your questions, which are not mutually exclusive :)

Yes, I would love to have reliable trimming options available to users although not as part of this specific WF here (which really tries to replicate the minion pipeline). What we could think of, besides having two WFs that need to be run sequentially, is offering a flavor of this WF, which includes preprocessing.

wm75:
I have pointed a few people working on long reads variant calling to our discussion here. Maybe they have something to say about what preprocessing steps to use.

mvdbeek:
> having two WFs that need to be run sequentially, is offering a flavor of this WF, which includes preprocessing.

This is a great example for subworkflows actually, and once that is in Galaxy, conditional subworkflow steps were users can choose if adapter trimming should be done.

mmiladi:
Hi @bwlang , I haven't seen any better (if at all) stand-alone ONT adapter trimmer than Porechop. Are you aware of any alternative?

A general clipper/trimmer is Nanofilt that might be helpful for some data: https://toolshed.g2.bx.psu.edu/repository?repository_id=b5209c19978dd767

This is from ARCTIC [preprint](https://www.biorxiv.org/content/10.1101/2020.09.04.283077v1.full):

>The results obtained with both porechop and Guppy with the --require_barcodes_both_ends argument were found to be comparable indicating MinKNOW ‘live basecalling’ can be used instead of porechop if this option is enabled. We also investigated single-end barcoding performance and found that although the LoCost protocol reduced misassignments (0.219%) they were still elevated with respect to any condition using double-ended barcoding. Comparing Guppy (3.2.10) which has been in production since the start of the outbreak to the most recent Guppy (4.0.14) we found stringency was increased with respect to the earlier version and produced results equivalent to porechop.

wm75:
Hmm @mmiladi, do you think it's worth then to have a Galaxy wrapper specifically for https://github.com/artic-network/Porechop instead of the officially unmaintained original?
Right now the two haven't diverged that dramatically and from the description major changes seem to be all about speed-ups. However, that's what they are using in the benchmarks in the preprint.

bwlang:
I love the subworkflow idea:  good to provide options where they make sense. In my experience, trimming adapters is a safe default, even if adapters have already been trimmed maybe you lose a few bases here and there, certainly not as harmful not trimming adapters with potential false positve outcomes.

I don't know of a galaxy wrapper for it, but bbduk could likely be tuned to work well for this application and will probably be much faster than porechop. If there is interest I could spend a bit of time to evaluate whether it could work - just let me know.

mmiladi:
@wm75 I don't have an answer for that. But in case that fork of porechop is the one ARTIC recommends then we should go for it. These options of the forked repo might influence the output...
```
`--barcode_headers`

This option writes the barcode calls to the header of the output fastq.

`--extended_headers`

This option also writes the barcode call for both ends of the read and the similarity scores for these matches

`--csv`

Writes a CSV file with one row per read giving the barcode call information
```

mmiladi:
It's a good idea to think about bbduk as an alternative! However imho I think that's something to discuss with/propose to the ARTIC group first. Their modification of porechop seems to have overcome its inefficiency. As bbduk isn't generally used on ONT data, using it in this Galaxy workflows would make the results a bit unicorn.

wm75:
@bwlang @mmiladi @NatPRoach I'm currently looking into the porechop Galaxy wrapper to see if it needs/could be improved before using it in a preprocessing step here.
In its current version this wrapper does *not* support *demultiplexing*, but only *trimming* of barcodes, *i.e.* it assumes that you have used an upstream tool to demultiplex based on the barcodes and are now feeding it reads from only one sample but with the barcode still present.

Is this a common scenario? We could split porechop into two tools maybe - a demultiplexer and a trimmer tool as suggested in the porechop [README](https://github.com/rrwick/Porechop#barcode-demultiplexing-1) itself.

mvdbeek:
> Is this a common scenario?

Seems unlikely iff the barcodes are used to demultiplex samples and we're looking at covid data, since otherwise a single run would contain multiple samples.

bwlang:
Even when demultiplexing, people have the option to keep the adapter sequences for later tools to take advantage of (at least on the gridion).  I can't say how common it might be that people use that option, but some people do keep those sequence routinely. If a scheme to trim adapters can be added I suspect it would be useful sometimes and not harmful other times (this can be tested).

bwlang:
IMHO, the demux is less important for this workflow.

NatPRoach:
What would really be ideal is if ONT would change the license on guppy so we could use the barcode / adapter trimming options in guppy (as well as the demux). I don't foresee that happening anytime soon however unfortunately.

PR #23, opened 2021-02-22
Initial version of sars-cov-2 variant-reporting workflow

mvdbeek:


mvdbeek:
This one is failing because planemo doesn't load the "built-in" tools, like `cat1`. So the question is, should we recommend TS tools or change planemo to load built-in tools ?

mvdbeek:
Although maybe that's not right, planemo should load distro tools: https://github.com/galaxyproject/planemo/blob/7f27dc0f9feca7ce172f1d0f9b32cd56c667607b/planemo/galaxy/config.py#L512
Guess we need a bit more logging ...

mvdbeek:
OK, for posterity, planemo was loading cat1, but only version 1.0.0, while the workflow specified version 2.0.0. Now 2.0.0 is specific to 21.01, while we were testing against master (which is release 20.09 currently). https://github.com/galaxyproject/planemo/pull/1146 also loads version 2.0.0 on newer Galaxy, but for now I've downgraded to 1.0.0 in https://github.com/galaxyproject/iwc/pull/23/commits/218f0be924189a64148a0fe0a0bcd8c83998623a, which is available in all galaxy instances, including 21.01 and newer

PR #27, opened 2021-03-18
Update SARS-CoV-2 ARTIC variation workflows

wm75:
- Introduces AmpliconBias INFO flag instead of AmpliconRemoval FILTER into Illumina version of the workflow.
- Changes the order of variant annotation and filtering in both WFs.


bwlang (approved):
looks fine to me
ga diff is surprisingly readable.

wm75:
> ga diff is surprisingly readable.

After quite some manual curation, I'm glad you noted it :)

bwlang:
> > ga diff is surprisingly readable.
> 
> After quite some manual curation, I'm glad you noted it :)

bummer that it will require manual effort -  i guess too much to hope for that such diffs would fall out naturally.

bwlang:
i feel a bit big for by britches in  merging this - but i guess I should....

mvdbeek:
Ah, I missed this. Feel free to ping me, and glad you've updated and merged it @wm75 @bwlang!

workflows/sars-cov-2-variant-calling/sars-cov-2-ont-artic-variant-calling/ont-artic-variation.ga
```diff
@@ -750,59 +750,77 @@
...
+            "inputs": [
+                {
+                    "description": "runtime parameter for tool SnpEff eff:",
+                    "name": "intervals"
+                },
+                {
+                    "description": "runtime parameter for tool SnpEff eff:",
+                    "name": "transcripts"
+                }
+            ],
```
mvdbeek, line 771: Is this correct @wm75 ? That looks like a bug we've fixed in https://github.com/galaxyproject/galaxy/pull/11652

workflows/sars-cov-2-variant-calling/sars-cov-2-ont-artic-variant-calling/ont-artic-variation.ga
```diff
@@ -888,135 +906,119 @@
...
+                    "output_name": "snpeff_output"
                 }
             },
-            "inputs": [],
+            "inputs": [
+                {
+                    "description": "runtime parameter for tool Lofreq filter",
+                    "name": "invcf"
+                }
+            ],
```
mvdbeek, line 923: And the same problem here ... unless this is intentional ?

workflows/sars-cov-2-variant-calling/sars-cov-2-ont-artic-variant-calling/ont-artic-variation.ga
wm75: This one looks correct. SnpEff really has these two optional runtime parameters. These are rather annoying because they prevent WFs with SnpEff from being runnable from the simplified WF run dialog, but the block existed also in the previous version (see the complementary diff further down), it's just that it shows up here as a change because of the altered tool order.

workflows/sars-cov-2-variant-calling/sars-cov-2-ont-artic-variant-calling/ont-artic-variation.ga
wm75: Hmm, this one might be wrong. Need to check.

wm75: Yeah, got introduced in this PR and seems wrong. The WF is running without issues though.
@mvdbeek Any specific problem to watch out for?

wm75:
Argh, going through the ONT WF again, I came across this rather expensive typo:

https://github.com/galaxyproject/iwc/blob/b7589068c146f1cd1374c922f16f43454cd7fe15/workflows/sars-cov-2-variant-calling/sars-cov-2-ont-artic-variant-calling/ont-artic-variation.ga#L42

mistyped `fastqsanger.gz` as `fastqsanger,gz`, which now causes non-sensical implicit conversions :(
It's better to leave off the format restriction entirely in this case?

mvdbeek:
I'll put it on the workflow WG roadmap ... yeah, in this case there's going to be automatic format restriction anyway, so 👍 for dropping the format restriction.

PR #28, opened 2021-03-23
Add ro-crate metadata

simleo:
Adds a script to generate [Workflow Testing RO-Crate](https://crs4.github.io/life_monitor/workflow_testing_ro_crate) metadata files that point to the existing GitHub testing workflow. [Support for monitoring GitHub Actions in Life Monitor](https://github.com/crs4/life_monitor/pull/117) is almost ready, but the expected values for `url` and `resource` should stay the same:

```json
"url": "https://api.github.com"
"resource": "repos/galaxyproject/iwc/actions/workflows/workflow_test.yml"
```

In general, for GitHub Actions, the expected `resource` pattern is:

```json
"repos/<ORG>/<REPO>/actions/workflows/<WORKFLOW>"
```

CC: @bgruening @ieguinoa @fbacall


fbacall:
I wonder if it would be better to only reference the files/folders that actually have metadata. i.e. remove README.md and .dockstore.yml etc.

It would be nice to keep it as concise as possible, so:
- Someone could potentially edit the metadata by hand instead of generating it using a script (like is done for `.dockstore.yml` : https://github.com/galaxyproject/iwc/blob/main/workflows/sars-cov-2-variant-calling/sars-cov-2-ont-artic-variant-calling/.dockstore.yml )
- It doesn't become out of date as soon as a file is added/removed/changed.

simleo:
`README.md` has to stay though, since Workflow Hub uses it for the workflow description.

simleo:
I pushed a new version that removes some files as suggested. It makes things a bit more awkward on the auto-generation side since more has to be assumed about the directory layout. Among things still open: adding a test *definition* that references the Planemo file, for a more complete testing ro-crate example (even though it's not supported by LM yet); the directory that contains test data should be named "test" according to the workflow ro-crate (and workflow testing ro-crate) spec.

simleo:
Good point @jmchilton, I wasn't sure whether the Galaxy team would be OK with renaming `test-data` to `test`. The idea we discussed with @fbacall was to have the `ro-crate-metadata.json` files already in place so that Workflow Hub could just zip the repos as ro-crates. Since the [testing crate spec](https://github.com/crs4/life_monitor/wiki/Workflow-Testing-RO-Crate) is currently based on the presence of a "test" directory, I proposed this setup, but now I see it clashes with an established convention.

The "test" name choice stems from the [workflow ro-crate spec](https://about.workflowhub.eu/Workflow-RO-Crate) which says the crate could contain a "test" dir for tests. When we started drafting the testing crate spec I guess we mostly had in mind a use case where a user can freely arrange the crate material to meet the format requirements. This is the first time we try to build testing crates around an existing setup, and I think it shows we could be more flexible. This also fits well with the fact that the monitoring scenario (i.e., monitoring of existing tests on a CI server) actually does not require a directory at all, but just references to appropriate URLs, so we should be able to model test suites even when no test directory is present.

Maybe this could be solved by having a context entity (e.g., TestSet) refer to the test suites with the `about` property, instead of the "test" data entity. References to test-related files, if present, could be handled by test definitions. But then the TestSet would be dangling. So I think it would be better to follow Stian's suggestion in https://github.com/ResearchObject/ro-crate/issues/122 and just add the `about` that refers to the test suites directly to the root entity. The test suites would certainly be "Contextual Entities important for this RO-Crate". @stain what do you think?

simleo:
Changed "test" back to "test-data" and updated RO-Crate metadata files to the latest [testing crate spec](https://github.com/crs4/life_monitor/wiki/Workflow-Testing-RO-Crate), which does not require the presence of a "test" directory anymore (see https://github.com/ResearchObject/ro-crate-py/pull/55).

simleo:
93fc6ef populates `workflows/README.md` with info on the directory structure and the RO-Crate metadata setup

simleo:
Updated description and removed draft status. Open for review!

workflows/README.md
```diff
@@ -1,2 +1,31 @@
...
+This directory contains all IWC workflows.
+
+The structure is as follows:
+
+* top-level directories represent categories, e.g., `sars-cov-2-variant-calling`;
+* directories right under the top level represent individual workflows, e.g., `sars-cov-2-consensus-from-variation`;
+* workflow directories contain:
+
+  * the `.ga` workflow file, e.g., `consensus-from-variation.ga`;
+  * a [Planemo](https://github.com/galaxyproject/planemo) test file, with the same name as the workflow file, but with a `.yml` extension, e.g., `consensus-from-variation.yml`;
```
mvdbeek, line 12:
```suggestion
  * a [Planemo](https://github.com/galaxyproject/planemo) test file, with the same name as the workflow file, but with a `-test.yml` extension, e.g., `consensus-from-variation-test.yml;
```

workflows/README.md
```diff
@@ -1,2 +1,31 @@
...
+  * a `test-data` directory with the test data used by Planemo;
+  * a [Dockstore](https://dockstore.org) [metadata file](https://docs.dockstore.org/en/develop/getting-started/github-apps/github-apps.html#workflow-yml-file) named `.dockstore.yml`;
+  * a [Workflow Testing RO-Crate](https://crs4.github.io/life_monitor/workflow_testing_ro_crate) metadata file named `ro-crate-metadata.json`;
+  * a `README.md` and a `CHANGELOG.md` file.
+
+
+## RO-Crate Metadata
```
mvdbeek, line 19: Is it OK for you if we generate the RO-Crate in the github workflow that deploys the repos to https://github.com/iwc-workflows ? I want to keep the artefacts to check in as minimal as possible.

workflows/gen_crates.py
```diff
@@ -0,0 +1,66 @@
+import json
+import os
+import pathlib
+
+# pip install 'rocrate==0.4.0'
+from rocrate.rocrate import ROCrate
+
+THIS_DIR = pathlib.Path(__file__).absolute().parent
+
+ROOT = THIS_DIR / "sars-cov-2-variant-calling"
```
mvdbeek, line 10: I guess we want to avoid hardcoding this ?

workflows/gen_crates.py
```diff
@@ -0,0 +1,66 @@
...
+
+GH_URL = "https://api.github.com"
+GH_RESOURCE = "repos/galaxyproject/iwc/actions/workflows/workflow_test.yml"
```
mvdbeek, line 13: What's the purpose of this ?

workflows/gen_crates.py
```diff
@@ -0,0 +1,66 @@
...
+        with open(wf_source) as f:
+            code = json.load(f)
+        author = [_["name"] for _ in code["creator"]]
+        if len(author) == 1:
+            author = author[0]
+        workflow = crate.add_workflow(wf_source, wf_id, main=True,
+                                      lang="galaxy", gen_cwl=False)
+        workflow["name"] = code["name"]
+        crate.root_dataset["author"] = author
+        wf_url = f"https://github.com/iwc-workflows/{entry.name}"
```
mvdbeek, line 50: Can we make everything that is a namespace configurable ? I am hoping this can become a template for other organizations as well.

workflows/sars-cov-2-variant-calling/sars-cov-2-consensus-from-variation/ro-crate-metadata.json
```diff
@@ -0,0 +1,141 @@
...
+            "resource": "https://w3id.org/ro/terms/test#resource",
+            "runsOn": "https://w3id.org/ro/terms/test#runsOn"
+        }
+    ],
+    "@graph": [
+        {
+            "@id": "./",
+            "@type": "Dataset",
+            "author": "Wolfgang Maier",
+            "datePublished": "2021-06-14T17:44:52.016833",
```
mvdbeek, line 25: Is there any fields that can list the specified version of the workflow ?

mvdbeek (commented):
This looks amazing, thanks so much @simleo!

workflows/README.md
```diff
@@ -1,2 +1,31 @@
-This contains all iwc worflows
-------------------------------
+# Workflows
+
+This directory contains all IWC workflows.
+
+The structure is as follows:
+
+* top-level directories represent categories, e.g., `sars-cov-2-variant-calling`;
+* directories right under the top level represent individual workflows, e.g., `sars-cov-2-consensus-from-variation`;
```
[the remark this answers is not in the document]
simleo: Done in b3aca89

workflows/README.md
simleo: Done in b3aca89

workflows/sars-cov-2-variant-calling/sars-cov-2-consensus-from-variation/ro-crate-metadata.json
simleo: Yes. A `version` property to be added to the `ComputationalWorkflow` entity. However, what would be the source for the property's value? I see version tags in the `CHANGELOG.md` files, but they're not machine-readable. They're listed in Markdown headings, but some use the atx style (leading hashes) while others use the setext style (underlining with equal signs and dashes). Moreover, some headings only contain the version tag (e.g., `## [0.1]`) while others also contain a date (e.g., `## [0.2] - 2021-04-30`).

We could try parsing the version tags with Python anyway, but that would not be very robust. Alternatives off the top of my head:

* add a `VERSION` file containing only the version string (old style)
* keep the version in the change log, but enforce a strict format for specifying it, e.g., `VERSION <VERSION_TAG>` on the first line or something like that
* keep things as they are now, but at least enforce the atx heading style and write down the convention somewhere as a guide for authors. Currently it seems to be `## [<VERSION>] <OTHER_METADATA>`, so the parsing code should do something like "search for the first level-2 heading, take the first non-whitespace character run and remove the leading and trailing square brackets"

workflows/gen_crates.py
simleo: These end up in the `TestInstance` entity as the `"url"` and `"resource"` properties:

```json
{
    "@id": "#test1_1",
    "@type": "TestInstance",
    "name": "test1_1",
    "runsOn": {
        "@id": "https://w3id.org/ro/terms/test#GithubService"
    },
    "url": "https://api.github.com",
    "resource": "repos/galaxyproject/iwc/actions/workflows/workflow_test.yml"
}
```

LifeMonitor [reads them](https://crs4.github.io/life_monitor/workflow_testing_ro_crate) to find out where to look for test build information.

workflows/sars-cov-2-variant-calling/sars-cov-2-consensus-from-variation/ro-crate-metadata.json
mvdbeek: There's a release field in the .ga files (which is manually maintained right now, but we're going to allow support for this in Galaxy). This is also what we use the create release tags for deployed repositories on github.

workflows/gen_crates.py
mvdbeek: Ah, excellent, that is very nice!

workflows/gen_crates.py
simleo: Done in 56191ba. Root directory is now a command line argument that can take 0 or more values, and defaults to the cwd.

workflows/sars-cov-2-variant-calling/sars-cov-2-consensus-from-variation/ro-crate-metadata.json
simleo: Excellent. Workflow version added in f1163a4

workflows/gen_crates.py
simleo: Done in 53495ce. I'd never thought about turning this quick hack into a proper tool before, but now I'm starting to see it doesn't take much to make it do something useful. In the LM project we're currently thinking about ways to make workflow testing easier in general, so this could even be the beginning of something. It's starting to shape up now. Thanks for the insight!

workflows/README.md
simleo: Sure. I've removed the RO-Crate metadata files in 2476874

simleo:
Thanks for the review @mvdbeek! I think all comments have been addressed now. This is the current status of the tool:

```
$ python gen_crates.py --help
usage: gen_crates.py [-h] [--owner STRING] [--repo STRING] [--workflow STRING]
                     [--target-owner STRING] [--planemo-version STRING]
                     [--zip-dir DIR_PATH] [--no-overwrite]
                     [ROOT_DIR [ROOT_DIR ...]]

Generate RO-Crate metadata for workflow repositories.

Assumes a two-tier directory structure where:

- the top-level directory represents a workflow category, e.g.,
  `sars-cov-2-variant-calling`;
- level 2 directories represent individual workflow repositories, e.g.,
  `sars-cov-2-consensus-from-variation`.

Level 2 directories are expected to contain:

- the `.ga` workflow file, e.g., `consensus-from-variation.ga`;
- a [Planemo](https://github.com/galaxyproject/planemo) test file with the
  same name as the workflow file, but with a `-test.yml` extension, e.g.,
  `consensus-from-variation-test.yml`;

positional arguments:
  ROOT_DIR              top-level directory

optional arguments:
  -h, --help            show this help message and exit
  --owner STRING        owner of the github workflow that runs the tests
  --repo STRING         repository that contains the github workflow
  --workflow STRING     github workflow file name (basename)
  --target-owner STRING
                        target github owner for repository deployment
  --planemo-version STRING
                        planemo version required to test the workflows
  --zip-dir DIR_PATH    create Workflow RO-Crate zip archives in this directory
  --no-overwrite        do not overwrite existing crates
```

mvdbeek (approved):
This is amazing! I'm gonna merge this and then open a PR to add the crates in the github workflow.

Issue #29, opened 2021-04-19
IWC goals 21.09


mvdbeek:
We've had a meeting today and we focused on these things we would like to accomplish:

- Get the IWC started
    - [ ] Writing how to contribute (@mvdbeek, @simonbray)
    - [ ] Best practices: (@mvdbeek) (Based on / linked to planemo documentation)
        - [ ] Don’t include internal galaxy version
        - [ ] Require release tag (or something)
    - [ ] Deposit workflows
        - [ ] OpenMS (@bernt-matthias )
        - [ ] RNAseq (@mvdbeek )
        - [ ] Update Covid Workflows (@wm75 )
        - [ ] Add some computational chemistry workflows (@simonbray )
    - [ ] Deploy workflows to usegalaxy.* (and interested community servers) (@mvdbeek)
        - [ ] And/or let deployers specify list of workflows to monitor
        - [ ] Maybe integration with usegalaxy-tools (???) (Stretchgoal)
    - [ ] Community outreach, find contributors (everyone)
   
- Enhance Galaxy for IWC
    - [ ] Link inside the app to update workflow (@jmchilton )
    - [ ] Update mechanism for TRS workflows (@jmchilton ) 
    - [ ] Display CHANGELOG/History 
    - [ ] Improve / fix TRS search 
    - [ ] “Request tools for this workflow” (Stretch goal)
    - [ ] Additional Markdown description field (@mvdbeek )
    - [ ] Prettier diffs (maybe in planemo)
        - [ ] Convert .ga files to format2 format (@ieguinoa )
        - [ ] Switch Export Default to YAML(?) (@jmchilton )
            - [ ] More testing? (@wm75 )
            - [ ] More manual inspection

- Continue improving command line execution with planemo:
    - [ ] Rerun with planemo run (@simonbray ) 
    - [ ] Workflow execution reports (@simonbray )

- Static Display of Workflows in Repo (Stretch Goal - @jmchilton )
    - [ ] Linking test data, pages for describing how to run.

nekrut:
Explore connection with the Cancer Informatics WG as they have a number of "production ready" workflows. There are also canter workflows being developed by Christophe Antoniewsky

PR #31, opened 2021-04-30
Add the covid-19 consensus workflow

wm75:
This is the last of the covid19.galaxyproject.org genomics workflows,
which still isn't deposited outside Galaxy.


wm75:
The version in this PR actually contains a couple of bugs and quirks, which I've fixed this week.
So once this gets merged I would open a new PR with the 0.2 release.

This one should be it's own version cause it's been used, e.g., to construct some of the COG-UK tracking project consensus sequences and we should have a proper release to refer to.

mvdbeek:
We need the `.dockstore.yml` file to trigger tests

workflows/sars-cov-2-variant-calling/sars-cov-2-consensus-from-variation/consensus.ga
```diff
@@ -0,0 +1,1735 @@
+{
```
mvdbeek, line 1: @wm75 I think this needs to be called consensus-from-variation.ga for planemo to pick up the test (or rename the test file to consensus-test.yml).

wm75: I see, thanks.

mvdbeek: Alright, that looks better:
```
Applying linter structure... WARNING
.. WARNING: Workflow contained output without a label
Applying linter tests... CHECK
.. CHECK: Tests appear structurally correct for workflows/sars-cov-2-variant-calling/sars-cov-2-consensus-from-variation/consensus-from-variation.ga
```

mvdbeek:
Need to install the missing tools on main ... one of the TODO items from #29 😆. I'm on it.

bwlang:
@wm75  this seems pretty complex... why not use ivar consensus?
https://andersen-lab.github.io/ivar/html/manualpage.html

wm75:
@bwlang with the collection of sars-cov-2-genomics WFs we are handling variant calling already (in a very reliable and sensitive way), and this WF here tries to build a consensus sequence from such a list of called variants.

ivar consensus is, of course, a much simpler solution, but it's also calling variants internally from its bam input again using samtools mpileup. So this is simply a different use case: if you want to get a fast consensus use ivar consensus, but we want a consensus sequence that incorporates the exact same set of variants that we called upstream.

In addition, most of the complexity in this WF comes from the aimed-for behavior described in the README. The core business of building the consensus FASTA is also just a single step (bcftools consensus) here. A bit of simplification will also follow in the first update of the WF.

That's the other important point: we need proper releases of this particular WF because just like the ARTIC PE variation WF and the reporting WF it's being used in our national viral genome surveillance tracking efforts (see https://usegalaxy-eu.github.io/posts/2021/04/29/sars-cov-2-monitoring/plain.html). The version in this PR has been run on ~ 35,000 COG-UK and a few Estonian samples already, and the updated version, which I would like to become iwc release 0.2, will be run on many more.

All this is not to say that we shouldn't have more (and possibly more lightweight) WFs for doing viral genomic analyses, and such WFs are still on my to do list. It's just that the ones we're already running on a lot of data are a priority for moving to iwc currently.

mvdbeek (approved):

PR #35, opened 2021-06-07
Fix sars-cov-2-pe-illumina-artic-variant-calling

drosofff:
- upgrading with with toolshed.g2.bx.psu.edu/repos/iuc/multiqc/multiqc/1.9+galaxy1


wm75:
-  Not hiding the json report may make things more consistent (at the expense of making the history look even more daunting to a novice user) -> I'm +-0 here. This output was hidden because the information in it gets visualized by multiqc anyway so normally there shouldn't be a need to inspect it.

- Updating multiqc is certainly good. May be worthwhile doing this in all variation WF flavors combined?

- "added explicitly input data ARTIC_nCoV-2019_v3 as a bed file, not as a bed6, which crashes the workflow"
I'm not seeing this in the diffs?

- "I also added the other immutable inputs NC_045512.2.fasta and ARTIC_amplicon_info_v3.tabular"
Good question whether such a thing should be done. Arguably, bundling the WF and its required resource files makes things more discoverable. OTOH it adds to the size of the repo.

drosofff:
> * Not hiding the json report may make things more consistent (at the expense of making the history look even more daunting to a novice user) -> I'm +-0 here. This output was hidden because the information in it gets visualized by multiqc anyway so normally there shouldn't be a need to inspect it.

I may be wrong but I though that hiding the json output messed up multiqc. In short multiqc was failing and I could only repair it by (i) unhiding the json fastp output and (ii) bumping the version of multiqc to 1.9+galaxy1

> * Updating multiqc is certainly good. May be worthwhile doing this in all variation WF flavors combined?

I'll do it certainly since I am preparing/checking all the workflows for the analysis of the french sequences

> * "added explicitly input data ARTIC_nCoV-2019_v3 as a bed file, not as a bed6, which crashes the workflow"
>   I'm not seeing this in the diffs?

Ah yeah, The annotation `BED file containing ARTIC primer positions. Can be retrieved from https://usegalaxy.eu/u/wolfgang-maier/h/covid-19-resources` is a bit misleading since the link most presumably points to your 2:ARTIC nCoV-2019 v3, whose datatype is bed6. if you just change it to bed, it's ok, but I started my test by doing a true bed6 to bed conversion, which does not work.

> * "I also added the other immutable inputs NC_045512.2.fasta and ARTIC_amplicon_info_v3.tabular"
>   Good question whether such a thing should be done. Arguably, bundling the WF and its required resource files makes things more discoverable. OTOH it adds to the size of the repo.

What I did not catch yet is how the actions test is done: where are stored the input data if not in the test-data folders ?

Anyway the PR is not ready, I have still some fixes to apply

wm75:
> > * "added explicitly input data ARTIC_nCoV-2019_v3 as a bed file, not as a bed6, which crashes the workflow"
> >   I'm not seeing this in the diffs?
> 
> Ah yeah, The annotation `BED file containing ARTIC primer positions. Can be retrieved from https://usegalaxy.eu/u/wolfgang-maier/h/covid-19-resources` is a bit misleading since the link most presumably points to your 2:ARTIC nCoV-2019 v3, whose datatype is bed6. if you just change it to bed, it's ok, but I started my test by doing a true bed6 to bed conversion, which does not work.
> 

I see, thanks for the explanation! Yes, the dataset in my shared resources history has it's datatype set to bed6, which indeed causes Galaxy's sniffer to complain:

```
Warning: The file 'Type' was set to 'bed6' but the file does not appear to be of that type
```
The reason is that the "score" column (column number 5) is not set to an integer as required in bed6. This is an annoying incompatibility in the bed file provided by the ARTIC network, from where that file was first obtained. If you were using `ivar trim` on the command line with that file it would complain about the wrong formatting and not trim any reads at all.
Our Galaxy wrapper is smart enough to change column 5 on the fly if it doesn't hold integer values - allowing users to work directly with the ARTIC network file.
The workflow specifies just `bed` as the input format for that file so it will accept the file no matter wether you set its datatype to `bed` or `bed6`. As long as you just use the file as it is, there shouldn't be any errors.

If you think it helps avoid confusion, I can try to change the datatype of the file in the shared history.

wm75:
> What I did not catch yet is how the actions test is done: where are stored the input data if not in the test-data folders ?

The tests are specified in `pe-artic-variation-test.yml`, which provides links to publically accessible files. That's the "standard" way of separating the WFs in this repo from the data they need to be tested.
In the specific case here, some of the test files also happen to be the most likely "real-world" input datasets so there is some appeal to bundling them with the WF, but I'm not fully convinced.

drosofff:
> The workflow specifies just `bed` as the input format for that file so it will accept the file no matter whether you set its datatype to `bed` or `bed6`. As long as you just use the file as it is, there shouldn't be any errors.

Ok I see.
So If I understand it is neither a bed nor a bed6. if galaxy does not complain with the bed datatype I think it would be less confusing to put it as a bed in your library. But it may be a matter of taste...

mvdbeek:
I agree with @wm75, the linked test data is preferable. If we can drop the the test data I think this looks fine, but we need to increase the release version in the .ga file (https://github.com/galaxyproject/iwc/blob/main/workflows/sars-cov-2-variant-calling/sars-cov-2-pe-illumina-artic-variant-calling/pe-artic-variation.ga#L14) and list the changes in https://github.com/galaxyproject/iwc/blob/main/workflows/sars-cov-2-variant-calling/sars-cov-2-pe-illumina-artic-variant-calling/CHANGELOG.md

drosofff:
At final, only upgrading multiqc and leaving the fastp json output hidden

workflows/sars-cov-2-variant-calling/sars-cov-2-pe-illumina-artic-variant-calling/pe-artic-variation.ga
```diff
@@ -312,15 +294,20 @@
...
                 {
                     "label": "fastp_html_report",
                     "output_name": "report_html",
-                    "uuid": "085c3083-2258-4673-90b6-22fe095cd4ad"
+                    "uuid": "341566b1-181a-4bb0-9d0f-37db45003b22"
+                },
+                {
+                    "label": "fastp_json_report",
+                    "output_name": "report_json",
+                    "uuid": "04955ac5-df51-41ff-bd19-a0a99c7ef761"
```
wm75, line 305: @mvdbeek I didn't even know this is possible. This adds the json report of fastp to the workflow outputs, but above the post job action of hiding that dataset is still defined. So these two things *can* be decoupled safely (just not through the GUI)?

mvdbeek: Yep, I thought that's possible

mvdbeek (approved):

workflows/sars-cov-2-variant-calling/sars-cov-2-pe-illumina-artic-variant-calling/CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+## [0.4] - 2021-06-16
+
+### Changed
+
+- upgrade multiqc to 1.9+galaxy1
```
mvdbeek, line 7:
```suggestion
- Upgrade multiqc to 1.9+galaxy1
```

workflows/sars-cov-2-variant-calling/sars-cov-2-pe-illumina-artic-variant-calling/pe-artic-variation.ga
wm75: It seems to work. The WF editor shows it as a workflow output, but when running the WF the dataset collection gets hidden. Nice!

wm75: I often edit the file directly to produce nicer diffs here. It's maybe not even more dangerous than using the WF editor, which tends to drop different things in different Galaxy versions ;)

mvdbeek:
Thanks a lot @drosofff @wm75!

PR #41, opened 2021-06-23
Add "Adding workflows" section, improve main readme

mvdbeek:
I've written this up while creating https://github.com/galaxyproject/iwc/pull/40, so this should include all steps


README.md
```diff
@@ -1,26 +1,34 @@
...
-
-### Meeting minutes:
- * [the notes](https://github.com/galaxyproject/iwc/issues/3) from the initial meeting at GCCBOSC2018 CoFest
+[![Galaxy Workflow Linting and Tests](https://github.com/galaxyproject/iwc/actions/workflows/workflow_test.yml/badge.svg?branch=main)](https://github.com/galaxyproject/iwc/actions/workflows/workflow_test.yml/badge.svg?branch=main)
+
+## The IWC maintains high-quality Galaxy Workflows
+
+Workflows are categorized in the [workflows](workflows/) directory, and listed in [Dockstore](https://dockstore.org/organizations/iwc) and [WorkflowHub](https://workflowhub.eu/search?utf8=%E2%9C%93&q=IWC#workflows).
+
+All workflows are reviewed and tested before and after publication, follow [best practices](https://planemo.readthedocs.io/en/latest/best_practices_workflows.html) and contain explicit versions and rich metadata, including:
```
bernt-matthias, line 9: Why "after publication"?

Maybe repace "and contain" by ", contain"

README.md
```diff
@@ -1,26 +1,34 @@
...
+- Author
+- Institutes
+
+Additionally the IWC will collect further best practices, tips and tricks, FAQs and assist the community in designing high-quality Galaxy workflows.
+
+## Importing Workflows into Galaxy
+
+To import IWC workflows into your Galaxy instance use the TRS workflow search in the Galaxy interface.
+Click on "Workflows", "Import" and click on the "search form" link.
+![Workflow Import GIF](static/iwc-import.gif)
```
bernt-matthias, line 21: The gif is cool. Would be even better if the workflows that are shown as result of the search would have a description.

bernt-matthias (approved):
Looks great

README.md
mvdbeek: We're going to test those workflows regularly, but let me make that more specific

README.md
mvdbeek: Yeah, that's a problem with gxformat2, I think. I'll update the gif once that's done.

README.md
```diff
@@ -1,26 +1,34 @@
...
+Anyone can contribute a Galaxy Workflow. Please check out the [Adding workflow guidelines](workflows/README.md#adding-workflows)
+If linting passes, tests pass, and human review passes, the PR is merged and
+
+- uploaded to a separate repository in <https://github.com/iwc-workflows> for versioning purposes
+- registered in the [iwc organization](https://dockstore.org/organizations/iwc) on [Dockstore](https://dockstore.org/)
+- registered in [WorkflowHub](https://workflowhub.eu/projects/33#workflows)
+
+## Becoming a IWC member
+
+Everyone is welcome and can help out with reviewing or creating workflows. Post a comment [here](https://github.com/galaxyproject/iwc/issues/4) with your expertise and we will add you to the IWC organization.
```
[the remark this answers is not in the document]
mvdbeek: Right, I'll rewrite that to say `Everyone is welcome and can help out with reviewing workflows.`

wm75:
Really nice!

In line 12 of the workflows/ README, you could maybe link directly to https://planemo.readthedocs.io/en/latest/test_format.html instead of to the planemo github repo?

wm75:
In the ro-crate section, it's a bit unclear whom the script is for, i.e. who's supposed to generate the data?

mvdbeek:
Yeah, let me move that into the deployment process, we can autogenerate this. The fewer things people need to submit the better IMO. We could probably even autogenerate the .dockstore.yml file, but that is a shortcut for planemo to find directories that should become a repository.

README.md
wm75: Interestingly, the organization column also doesn't get populated, but *this* info is available.

mvdbeek: Yeah, the vue component needs a little kick :)

mvdbeek: Also https://github.com/seek4science/seek/issues/637 would give us a chance to explore organizations. If you want to thumb up that issue maybe it'll move up the priorities ? 😄

wm75: @OlegZharkov maybe?

README.md
```diff
@@ -1,26 +1,37 @@
...
+- Author
+- Institutes
+
+Additionally the IWC will collect further best practices, tips and tricks, FAQs and assist the community in designing high-quality Galaxy workflows.
+
+## Importing Workflows into Galaxy
+
+To import IWC workflows into your Galaxy instance use the TRS workflow search in the Galaxy interface.
+Click on "Workflows", "Import" and click on the "search form" link.
+![Workflow Import GIF](static/iwc-import.gif)
```
wm75, line 23: Can you add a note under that animation saying: when searching against workflowhub.eu, please use organization:"IWC Workflows" as the search term?

mvdbeek: This needs to be fixed on the workflowhub side. iwc is the right organization name.

mvdbeek: Do you have a workflowhub account ? I do, but I can't log in, hanging in a loop. If you do, maybe you could submit the form https://workflowhub.eu/home/feedback ?

mvdbeek: Finally managed it in a private session.
Screenshot 2021-06-24 at 16 37 26

wm75: Well, yes and no:

1. I don't quite understand what dockstore is doing either: when you look at the WFs returned by the organization:iwc query, they all list "iwc-workflows" as the organization (though undoubtedly we're iwc on the dockstore website). Not sure whether the query only works accidentally (organization:workflows seems to be working just as well).
2. Until this is unified upstream, we might still want to provide working instructions here (and simplify it later)?

mvdbeek: https://dockstore.org/organizations/iwc -> that is the organization name.
From the API:
```json
  {
    "aliases": {},
    "avatarUrl": "https://galaxyproject.org/images/galaxy-logos/galaxy_project_logo_square.png",
    "dbCreateDate": 1599229556066,
    "dbUpdateDate": 1624453421503,
    "description": "# IWC - Intergalactic Workflow Commission\n\n### Reproducibility is important. Our goals are to:\n - foster workflow use\n - incorporate versioning\n - capture more metadata: (names, versions, authors, use cases, etc.)\n - help scientists find workflows!",
    "displayName": "The Galaxy Intergalactic Workflow Comission",
    "email": "iwc@galaxyproject.org",
    "id": 33,
    "link": "https://github.com/galaxyproject/iwc",
    "location": null,
    "name": "iwc",
    "starredUsers": [
      {
        "id": 3
      },
      {
        "id": 47541
      },
      {
        "id": 53591
      }
    ],
    "status": "APPROVED",
    "topic": "This group exists to create and curate a registry of best practice community workflows for Galaxy.",
    "users": null
  },
```

mvdbeek: I think iwc-workflows is taken from the github namespace ... while I don't love it that makes sense.

mvdbeek: Reported that here: https://discuss.dockstore.org/t/organization-mismatch-if-github-app-installed-in-separate-namespace/4400.

I guess it doesn't hurt to add a note.

mvdbeek: Alright, with Denis' response I looked a bit closer, and searching for `iwc-workflows` should also return our workflows. So https://github.com/galaxyproject/galaxy/pull/12196 fixes that in Galaxy. So then I guess it's somewhat consistent. I updated the gif too.

mvdbeek:
> Would be even better if the workflows that are shown as result of the search would have a description

Hope that'll be fixed with https://github.com/galaxyproject/dockstore-galaxy-interface/pull/19

README.md
drosofff: For information, when I follow exactly the gif instructions (Galaxy search form), the search works only with "iuc" and fails if I add "-workflows" or anything after "iuc"

mvdbeek: Thanks, yeah, that will be fixed in https://github.com/galaxyproject/galaxy/pull/12196

wm75: @drosofff quoting the organization name is what's working right now, i.e. `organization:"iwc-workflows"` for dockstore and `organization:"IWC Workflows"` for workflowhub.

README.md
```diff
@@ -1,26 +1,40 @@
...
+
+When importing a workflow form workflowhub.eu you currently need to search for `organization: "IWC Workflows"`.
```
wm75, line 26:
```suggestion
When importing a workflow form workflowhub.eu you currently need to search for `organization:"IWC Workflows"`.
```

a space between the key and the value seems to break things, too.

mvdbeek: Ugh ... we really need to have that search interface you suggested.

README.md
mvdbeek: I've deployed the update to usegalaxy.org, so organization:iwc-workflows works there.

README.md
mvdbeek: Quick fix is https://github.com/galaxyproject/galaxy/pull/12197, then we also don't need to quotes anymore

drosofff: only for 21.05 I guess ;-)

mvdbeek: We can backport this a bit further when it has stabilised 😆

mvdbeek:
I'm gonna go with @bernt-matthias' approval :). We can (and should) revisit this whenever we have updates. Thanks for the reviews @bernt-matthias @wm75 @drosofff!

PR #42, opened 2021-06-29
Fix title in README for sars-cov-2-se-illumina-wgs-variant-calling

bebatut:
to fix the description on [WorkflowHub](https://workflowhub.eu/workflows/112) which is inconsistent ("COVID-19: variation analysis on WGS PE data" for the SE workflow)


wm75 (approved):
Thanks for spotting this @bebatut! Embarassing mistake.

mvdbeek:
Without a bump in the `release` key of the workflow file we won't create a new release / update https://github.com/iwc-workflows/sars-cov-2-se-illumina-wgs-variant-calling. Is that OK or do we want a new release ?

wm75:
I don't know, honestly. It seems like a ridiculously small change for a release (and bumping the release in the workflow kind of suggests that the workflow has changed). Maybe it's ok to know that the next release will carry the fixed README?

PR #49, opened 2021-07-19
Update gen_crates.py

simleo:
Updates `gen_crates.py` to handle new situations brought about by the addition of the `parallel-accession-download` workflow:

* Use [the same logic as Planemo](https://github.com/galaxyproject/planemo/blob/0.74.5/planemo/runnable.py#L480) to find the test file. Now the script finds `data-fetching/parallel-accession-download/parallel-accession-download-tests.yml`
* Handle `Organization` creators


mvdbeek (approved):

mvdbeek:
Thanks @simleo!

PR #51, opened 2021-07-21
Generate RO-Crate metadata before deployment

simleo:
Updates the GitHub workflow with steps to generate RO-Crate metadata for all the workflows before deployment, using the `gen_crates.py` tool.

The deployment phase will actually be skipped for _this_ PR, since no workflow has changed. I'm not sure what would be the best way to "force" the addition of the RO-Crate files to the repos, since the deploy action updates a repo only for new workflow releases.


mvdbeek (approved):
Thanks @simleo! There's no way but to bump the individual workflows. This can be a patch version bump, and the changelog can mention that there is no functional change. I think a script will be important to do this in the future.

The PR looks good for now, but I'm playing with the idea of moving the crate generation into a separate github action, or to move it into planemo itself, so people can use it outside of the iwc too.

PR #37, opened 2021-06-11
Fix reference to test file in sars-cov-2-se-illumina-wgs-variant-calling

simleo:
Fixes a problem that breaks the build (see https://github.com/galaxyproject/iwc/runs/2803865520) now that `planemo-ci-action` has been updated to include https://github.com/galaxyproject/planemo-ci-action/pull/19.

Before this fix:
```
$ planemo workflow_lint --fail_level error workflows/sars-cov-2-variant-calling/sars-cov-2-se-illumina-wgs-variant-calling
Applying linter dockstore... FAIL
.. ERROR: .dockstore.yml workflow entry references absent file /se-wgs-variation.ga-test.yml
Applying linter tests... CHECK
.. CHECK: Tests appear structurally correct for workflows/sars-cov-2-variant-calling/sars-cov-2-se-illumina-wgs-variant-calling/se-wgs-variation.ga
```

After this fix:
```
(venv) [simleo@neuron:iwc (fix_dockstore)]$ planemo workflow_lint --fail_level error workflows/sars-cov-2-variant-calling/sars-cov-2-se-illumina-wgs-variant-calling
Applying linter tests... CHECK
.. CHECK: Tests appear structurally correct for workflows/sars-cov-2-variant-calling/sars-cov-2-se-illumina-wgs-variant-calling/se-wgs-variation.ga
```

PR #52, opened 2021-07-23
Bump all versions

simleo:
See https://github.com/galaxyproject/iwc/pull/51#pullrequestreview-711903044.

* Bumps the patch version for all workflows, mentioning the addition of the RO-Crate metadata files (which will hopefully be generated by the GitHub workflow upon deployment) in the logs.
* Adds a new Python tool that was used to perform the above bump:

```
python bump_version.py -t Added -m "Added RO-Crate metadata file. No functional changes."
```

Note that the tool has also been used to record the change to `sars-cov-2-se-illumina-wgs-variant-calling` done in #37, which was still unrecorded:

```
python bump_version.py workflows/sars-cov-2-variant-calling/sars-cov-2-se-illumina-wgs-variant-calling -v 0.1.1 -m "Fix reference to test file in .dockstore.yml" -d 2021-06-14 -t Fixed
```

More details for the various options:

```
$ python bump_version.py --help
usage: bump_version.py [-h] [--exclude [PATH [PATH ...]]] [-v STRING]
                       [-m STRING] [-d STRING] [-t Added|Changed|Fixed]
                       [-b MAJOR|MINOR|MICRO]
                       [ROOT_DIR [ROOT_DIR ...]]

Bump workflow repository version.

Workflow repositories are searched for starting from the specified root
directories (the default is to search below the current directory). Uses the
same searching logic and definition of repository as the ci_find_repos Planemo
command (any directory with a .shed.yml or .dockstore.yml file).

Workflow repositories are expected to contain:

- the .ga workflow file (e.g., "consensus-from-variation.ga"), which is
  expected to list its version in the "release" metadata field
- a CHANGELOG.md file with atx-style Markdown headers (the ones with the
  leading hash marks), with the convention that:
  - the top-level header is "# Changelog"
  - level 2 headers start with ## <VERSION>

Example:

# Changelog

## [0.4] - 2021-06-16

### Changed

- Upgrade multiqc to 1.9+galaxy1

positional arguments:
  ROOT_DIR              top-level directory

optional arguments:
  -h, --help            show this help message and exit
  --exclude [PATH [PATH ...]]
                        paths to exclude while searching for workflow repos
  -v STRING, --version STRING
                        set new workflow version to this value (ignores '-b')
  -m STRING, --msg STRING
                        log message
  -d STRING, --date STRING
                        log date as YYYY-MM-DD
  -t Added|Changed|Fixed, --entry-type Added|Changed|Fixed
                        log entry type
  -b MAJOR|MINOR|MICRO, --bump MAJOR|MINOR|MICRO
                        version part to bump
```


mvdbeek (approved):
Brilliant, very useful. Thanks so much @simleo!

simleo:
[It worked](https://github.com/iwc-workflows/parallel-accession-download/blob/main/ro-crate-metadata.json) :partying_face:

PR #53, opened 2021-07-29
RO-Crate upload to WorkflowHub

simleo:
Extends the `gen_crates` tool with functionality to upload RO-Crates to WorkflowHub.

The tool needed some additional info to perform this action:

* WorkflowHub URL: defaults to https://workflowhub.eu, can be overridden with `--hub-url` (can point to https://dev.workflowhub.eu or http://localhost:3000 for testing)
* WorkflowHub API key for authentication: passed via the `HUB_API_KEY` environment variable (passing this via CLI would not be safe). We should be able to set this from a secret in GitHub Actions
* WorkflowHub team/project name: read from `.workflowhub.yml`, can be overridden with `--hub-project`. Can be omitted in `.workflowhub.yml`, but then `--hub-project` must be specified.
* WorkflowHub workflow name: read from `.workflowhub.yml`. If a workflow with this name is found in the given project, the tool creates a new version for it; if not, it creates a new workflow. In any event, the tool prints the resulting WorkflowHub link, which provides the final workflow id and version

Example:

```console
$ # Workflows not in WorkflowHub yet
$ python gen_crates.py --upload --hub-url https://dev.workflowhub.eu
processing data-fetching/parallel-accession-download
  archived as /tmp/iwc_7zp3847w/parallel-accession-download.crate.zip
  uploaded as https://dev.workflowhub.eu/workflows/193?version=1
processing sars-cov-2-variant-calling/sars-cov-2-consensus-from-variation
  archived as /tmp/iwc_7zp3847w/sars-cov-2-consensus-from-variation.crate.zip
  uploaded as https://dev.workflowhub.eu/workflows/194?version=1
$ # Now all workflows have been registered
$ python gen_crates.py --upload --hub-url https://dev.workflowhub.eu
processing data-fetching/parallel-accession-download
  archived as /tmp/iwc_6dds16hx/parallel-accession-download.crate.zip
  uploaded as https://dev.workflowhub.eu/workflows/193?version=2
processing sars-cov-2-variant-calling/sars-cov-2-consensus-from-variation
  archived as /tmp/iwc_6dds16hx/sars-cov-2-consensus-from-variation.crate.zip
  uploaded as https://dev.workflowhub.eu/workflows/194?version=2
```

The `.workflowhub.yml` file format should be self-explanatory. Example:

```yaml
version: '0.1'
registries:
- url: https://workflowhub.eu
  project: iwc
  workflow: sars-cov-2-consensus-from-variation/COVID-19-CONSENSUS-CONSTRUCTION
- url: https://dev.workflowhub.eu
  project: LifeMonitorDev
  workflow: sars-cov-2-consensus-from-variation/COVID-19-CONSENSUS-CONSTRUCTION
```

We should be able to integrate this in the GitHub Actions workflow, so that (scientific) workflows are automatically uploaded to WorkflowHub at each release.

* An IWC admin should add a secret containing their WorkflowHub API key to this repository
* A WorkflowHub admin should add that IWC admin to the "iwc" team on WorkflowHub, so that the above API key would allow the workflow upload

### DIFF
--- a/workflows/gen_crates.py
+++ b/workflows/gen_crates.py
@@ -66,6 +66,15 @@ HUB_URL = "https://workflowhub.eu"
 HUB_API_KEY = os.getenv("HUB_API_KEY")
 HUB_CFG_NAME = ".workflowhub.yml"
 HUB_CFG_VERSION = "0.1"
+HUB_API_HTTP_HEADERS = {
+    "Content-type": "application/vnd.api+json",
+    "Accept": "application/vnd.api+json",
+    "Accept-Charset": "ISO-8859-1",
+}
+DOCKSTORE_CFG_NAME = ".dockstore.yml"
+DOCKSTORE_CFG_VERSION = "1.2"
+
+_WF_CACHE = {}
 
 
 def get_wf_id(crate_dir):
@@ -111,6 +120,18 @@ def handle_creator(ga_json, crate, workflow):
         workflow["creator"] = ro_creators
 
 
+def get_workflow_name(repo_dir):
+    repo_dir = Path(repo_dir)
+    cfg_path = repo_dir / DOCKSTORE_CFG_NAME
+    if not cfg_path.is_file():
+        raise RuntimeError(f"{cfg_path} not found")
+    with open(cfg_path, "rt") as f:
+        cfg = yaml.load(f, Loader=Loader)
+    assert str(cfg["version"]) == DOCKSTORE_CFG_VERSION
+    wf_entry = cfg["workflows"][0]  # assuming first wf is the main wf
+    return f"{repo_dir.name}/{wf_entry['name']}"
+
+
 def make_crate(crate_dir, target_owner, resource, planemo_version):
     wf_id = get_wf_id(crate_dir)
     planemo_id, planemo_source = get_planemo_id(crate_dir, wf_id)
@@ -121,7 +142,7 @@ def make_crate(crate_dir, target_owner, resource, planemo_version):
     workflow = crate.add_workflow(wf_source, wf_id, main=True,
                                   lang="galaxy", gen_cwl=False)
     handle_creator(code, crate, workflow)
-    workflow["name"] = code.get("name", crate_dir.name)
+    workflow["name"] = get_workflow_name(crate_dir)
     try:
         workflow["version"] = code["release"]
     except KeyError:
@@ -156,7 +177,7 @@ def find_repos(paths, exclude=()):
     return [Path(_) for _ in filter_paths(ctx, raw_repos, path_type="repo", **kwargs)]
 
 
-def upload_crate(crate_archive, proj_id, hub_url=HUB_URL, wf_id=None):
+def upload_crate(crate_archive, proj_id, hub_url=HUB_URL, wf_id=None, wf_name=None):
     if not HUB_API_KEY:
         raise RuntimeError("API key not found. Set the HUB_API_KEY environment variable")
     endpoint = f"{hub_url}/workflows"
@@ -170,15 +191,25 @@ def upload_crate(crate_archive, proj_id, hub_url=HUB_URL, wf_id=None):
         }
         r = requests.post(endpoint, headers=headers, files=payload)
     r.raise_for_status()
+    # update name
+    headers.update(HUB_API_HTTP_HEADERS)
+    id_ = r.json()["data"]["id"]
+    if wf_name and wf_name != r.json()["data"]["attributes"]["title"]:
+        payload = {
+            "data": {
+                "id": id_,
+                "type": "workflows",
+                "attributes": {
+                    "title": wf_name
+                }
+            }
+        }
+        r = requests.patch(f"{hub_url}/workflows/{id_}", headers=headers, json=payload)
+        r.raise_for_status()
     if wf_id:
         return r.json()
     # new workflow - update access policy
     wf_id = r.json()["data"]["id"]
-    headers.update({
-        "Content-type": "application/vnd.api+json",
-        "Accept": "application/vnd.api+json",
-        "Accept-Charset": "ISO-8859-1",
-    })
     payload = {
         "data": {
             "id": wf_id,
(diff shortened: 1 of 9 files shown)


fbacall:
Does this write the workflow ID back to the config file when a new workflow is created?

fbacall:
Sharing a single API key across multiple WorkflowHub instances would not work.

workflows/sars-cov-2-variant-calling/sars-cov-2-consensus-from-variation/.workflowhub.yml
```diff
@@ -0,0 +1,5 @@
+version: "0.1"
+registries:
+- url: "https://workflowhub.eu"
+  project: 33
```
mvdbeek, line 4: Can we work on this ? These numeric ids are really unfortunate.

simleo:
> Sharing a single API key across multiple WorkflowHub instances would not work.

Sure, the idea is to have the GitHub workflow only upload to https://workflowhub.eu. The config file supports other registry instances to enable manual testing.

workflows/sars-cov-2-variant-calling/sars-cov-2-consensus-from-variation/.workflowhub.yml
simleo: For a while I've played with the idea of allowing names instead of ids for the project and workflow. For projects it's fairly straightforward: just get `/projects` then select by `title`, which is a unique attribute and it's set once and for all ("iwc" in https://workflowhub.eu). For workflows, however, `title` is not unique; moreover, the workflow's name could change in the course of its revision history. This means we might end up having multiple instances of the same workflow if we're not careful.

My proposal to make this manageable is to **set the workflow's name/title to its repository name** (e.g., `parallel-accession-download`, `sars-cov-2-consensus-from-variation`), which is not going to change and should stay unique within the iwc project (unless we mess things up and fail to request a new version somehow).

This involves selecting all workflows that belong to a project, but I had a technical problem while trying that: the `'relationships'` field in the project structure returned by the API does not list workflows (I think it's [this issue](https://github.com/seek4science/seek/issues/447)), so I had to build a reverse map:

```python
def get(endpoint):
    r = session.get(base_url + endpoint)
    r.raise_for_status()
    return r.json()['data']

proj_to_wf = {}
workflows = get("/workflows")
for stub in workflows:
    wf = get(f"/workflows/{stub['id']}")
    for r in wf["relationships"]["projects"]["data"]:
        proj_to_wf.setdefault(r["id"], set()).add(wf["id"])
```

That solves the problem but it's horribly expensive since it requires to individually get all workflows. @fbacall can you help with this?

fbacall: Try again, it should list them now

fbacall: BTW, re: workflow naming, see: iwc-workflows/parallel-accession-download#1

simleo:
Now both projects and workflows can be specified by name. Workflow names are set as per iwc-workflows/parallel-accession-download#1.

simleo:
> * An IWC admin should add a secret containing their WorkflowHub API key to this repository
> * A WorkflowHub admin should add that IWC admin to the "iwc" team on WorkflowHub, so that the above API key would allow the workflow upload

@mvdbeek @fbacall Do you think the PR is good to go now? If so, I'd like to merge it. Then, if you can perform the above steps, we'll be able to update the GH workflow and set up automatic upload.

The current IWC workflows will have to drop the version bit from their name before this can work (e.g., [sars-cov-2-consensus-from-variation/COVID-19-CONSENSUS-CONSTRUCTION (v0.2.1)](https://workflowhub.eu/workflows/138) needs to change to just "sars-cov-2-consensus-from-variation/COVID-19-CONSENSUS-CONSTRUCTION".

workflows/gen_crates.py
```diff
@@ -56,6 +62,113 @@
 PLANEMO_VERSION = f">={planemo.__version__}"
 PLANEMO_TEST_SUFFIXES = ["-tests", "_tests", "-test", "_test"]
 PLANEMO_TEST_EXTENSIONS = [".yml", ".yaml", ".json"]
+HUB_URL = "https://workflowhub.eu"
+HUB_API_KEY = os.getenv("HUB_API_KEY")
```
mvdbeek, line 66: Can we do this without the API key and pushing from our side? I'd prefer a github app like we've installed for dockstore, that would make the repository more portable across organizations or owners.

mvdbeek (commented):
> The current IWC workflows will have to drop the version bit from their name before this can work (e.g., [sars-cov-2-consensus-from-variation/COVID-19-CONSENSUS-CONSTRUCTION (v0.2.1)](https://workflowhub.eu/workflows/138) needs to change to just "sars-cov-2-consensus-from-variation/COVID-19-CONSENSUS-CONSTRUCTION".

👍 I think that is a good change

How are we going to maintain versions? If we have to trigger the upload (I'd prefer if we didn't have to) I think the CI should be running from the deployed repo's release events, I think we shouldn't use the central repo for this.

> ```
> processing sars-cov-2-variant-calling/sars-cov-2-consensus-from-variation
>   archived as /tmp/iwc_6dds16hx/sars-cov-2-consensus-from-variation.crate.zip
>   uploaded as https://dev.workflowhub.eu/workflows/194?version=2
> ```

That is just an example, right ? The actual version would be `https://dev.workflowhub.eu/workflows/194?version=0.2` or whatever version we define ?

simleo:
> > ```
> > processing sars-cov-2-variant-calling/sars-cov-2-consensus-from-variation
> >   archived as /tmp/iwc_6dds16hx/sars-cov-2-consensus-from-variation.crate.zip
> >   uploaded as https://dev.workflowhub.eu/workflows/194?version=2
> > ```
> 
> 
> That is just an example, right ? The actual version would be `https://dev.workflowhub.eu/workflows/194?version=0.2` or whatever version we define ?

Links to workflows in WorkflowHub use WorkflowHub's internal version IDs (auto-incrementing integers). I don't think these IDs can be mapped to user-defined version tags (e.g., to show up in the web app and in the API workflow object) at the moment, @fbacall correct me if I'm wrong.

In any event, the actual version tag as defined in IWC is available from the workflow's RO-Crate, since it's recorded in the metadata.

fbacall:
Yes, they are just auto-incrementing database IDs at the moment

mvdbeek (approved):
Let's get this in, I guess we can still work out the modalities later when we have some experience using this all.

simleo:
Thanks Marius! I have a couple of questions regarding your comments:

> How are we going to maintain versions? If we have to trigger the upload (I'd prefer if we didn't have to) I think the CI should be running from the deployed repo's release events, I think we shouldn't use the central repo for this.

The plan was to have the workflows automatically registered and subsequently updated at each deployment, so that @fbacall can shut down his daily "scraper" and we can have the workflows synced (to WorkflowHub and, at some point, to LifeMonitor) as soon as they are published. What would be so bad about triggering the upload from the central CI job?

> Can we do this without the API key and pushing from our side? I'd prefer a github app like we've installed for dockstore, that would make the repository more portable across organizations or owners.

I don't have experience with GitHub apps. What would this entail? Where should the app be hosted, and which org should own it? Could the existing code be reused, or would it need to be rewritten / adapted?

mvdbeek:
> What would be so bad about triggering the upload from the central CI job?

I feel like that is infrastructure we, or organizations that want to host workflows, shouldn't have to maintain. It also seems much easier to install an app as compared to setting up the secret. 
It would also complicate the central CI job somewhat, since it would have to check what releases exist in the deployed repo and (re)-publish everything that hasn't been published. The dockstore app is really great in that regard.

> I don't have experience with GitHub apps. What would this entail? Where should the app be hosted, and which org should own it? Could the existing code be reused, or would it need to be rewritten / adapted?

I have only built a github app using https://probot.github.io/ which was ... a little tricky, mostly because of my lack of typescript knowledge, but there also integrations in other language (though they are not documented as well). I suppose you could re-use some of the code if you use one of the integrations written in python (or use a shell to run `gen_crates`). One difference in the model is that the app would sync the workflows to workflowhub, so I guess the logic that authenticates a user/org would need to be couple to the repository, instead of using an API key.

PR #55, opened 2021-08-02
Makes it easier for users to update for longer amplicons

bwlang:


bwlang:
an alternate path:
https://usegalaxy.eu/u/brad_langhorst/w/covid-19-variation-analysis-of-artic-ont-data-automaxlen
let me know what you think about this approach for automatic max length calculation (im using max gap * 1.9 since that produces approximately 700 for ARTIC v3 and ~ 2200 for a long amplicon design (1.4 kb amplicons)

PR #59, opened 2021-09-17
ONT-Artic workflow fixes and enhancements

wm75:


wm75:
It's the first two `*` points from the changelog that prompted this update. The prior version just isn't that usable on large batches of real-world data.

Compared to just updating the medaka variant tool to its latest version and leaving everything else unaltered, this more complex version here clearly enhances sensitivity (tested on a batch of 220 samples).

wm75:
https://github.com/galaxyproject/iwc/pull/59/checks?check_run_id=3677521934#step:6:5357

Ah, of course, this WF needs version 1.3.2+galaxy1 of medaka variant, but was tested against +galaxy0.

mvdbeek:
I'll update main.

mvdbeek:
Alright, triggered a re-run

mvdbeek:
There's one linter warning, `WARNING: Workflow contained output without a label`, could check this one ?

wm75:
Aha, the warning came from the two min and max length workflow input parameters. The WF editor attaches workflow outputs to input params, which you cannot see or modify through the GUI. I removed those outputs now. Hope that's the correct way to go?

mvdbeek:
Hmm, these are actual workflow outputs. I think we may want to adjust the linter instead. Can you undo the last commit ? I'll  have to think a bit about this.

mvdbeek:
Thanks a lot @wm75!

PR #63, opened 2021-09-21
add report and fail level for lint

bernt-matthias:
values are the defaults that will be introduced in https://github.com/galaxyproject/planemo-ci-action/pull/23 but I think its a good idea to add them here anyway .. just in case anyone uses the repo as template.

fail level warn will be more strict that before, but I guess this is a good
idea

### DIFF
--- a/.github/workflows/workflow_test.yml
+++ b/.github/workflows/workflow_test.yml
@@ -92,6 +92,8 @@ jobs:
       with:
         mode: lint
         workflows: true
+        report-level: all
+        fail-level: warn
         repository-list: "${{ needs.setup.outputs.repository-list }}"
 
   test:


mvdbeek:
Before we can do this we need to fix the warnings, some of which need some thinking about (we warn on unlabelled workflow outputs for input parameters ... which kind of makes sense, but I don't know how to handle this)

wm75:
Would it make sense to just auto-label such outputs with the input param name?

mvdbeek:
I don't know. You mean the step label, right ? That wouldn't work for parameters that produce more than 1 output ... we don't have any of these, but you could imagine a range parameter, or an expression or regex that produces multiple values. Unless we'd always represent them as a single output ... which would make them less useful as input parameters, I guess.

We could auto-suggest something in the UI, I think that might be OK.
Maybe we just skip linting workflow outputs for input type parameters for now. @jmchilton do you have any thoughts on this ?

mvdbeek:
I think we need a new release of planemo-ci-action for this, right ?

bernt-matthias:
> I think we need a new release of planemo-ci-action for this, right ?

Indeed. Maybe including https://github.com/galaxyproject/planemo-ci-action/pull/20 as well?

PR #64, opened 2021-10-04
Expression Atlas Smartseq/Droplet downstream workflow

pcm32:
This PR aims to add the EMBL-EBI Single Cell Expression Atlas downstream analysis workflow as used to produce data with the resource, with a hopefully lightweight example.

I have a few concerns though:

1.- I would hope that the WorkflowHub.eu entry can be annotated to give credits to the person who wrote the workflow. I think that it will currently show as IWC, and while I of course value all the excellent feedback that the IWC community will provide, I still think that the main credit should go to the person writing the workflow (Jon Manning in this case).
2.- It would be great as well if, besides setting the author of the workflow to be passed to WorkflowHub.eu, additional organisations could be added (so that we could add, for instance, the EBI Gene Expression Team, next to IWC).
3.- Our workflow setup currently allows certains steps to fail. For instance, when it comes to the cluster phase, I think that the process relies on some convergence criteria, and for some resolutions value a dataset might not converge if I remember well. Will the testing here consider the workflow failed if one of those steps failed? On our setup (https://github.com/ebi-gene-expression-group/galaxy-workflow-executor) we allow certain steps to fail while still considering the workflow to be successful.

I would be happy to PR needed changes to deal with 1 and 2 if given some directions on where you would do these changes and how you would like them.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "workflows/single-cell-RNAseq/.ebi-gene-expression-group/scxa-workflows"]
+	path = workflows/single-cell-RNAseq/.ebi-gene-expression-group/scxa-workflows
+	url = https://github.com/ebi-gene-expression-group/scxa-workflows
--- a/workflows/single-cell-RNAseq/SingleCellExpressionAtlas-downstream/SingleCellExpressionAtlas-smartseq-test.yml
+++ b/workflows/single-cell-RNAseq/SingleCellExpressionAtlas-downstream/SingleCellExpressionAtlas-smartseq-test.yml
@@ -1,0 +1,17 @@
+- doc: Test Single Cell Expression Atlas smartseq execution using E-MTAB-6077
+  job:
+    genemeta:
+      class: File
+      location: http://ftp.ebi.ac.uk/pub/databases/microarray/data/atlas/sc_experiments/E-MTAB-6077/Danio_rerio.GRCz11.99.gtf.gz
+    genes:
+      class: File
+      location: http://ftp.ebi.ac.uk/pub/databases/microarray/data/atlas/sc_experiments/E-MTAB-6077/E-MTAB-6077.aggregated_counts.mtx_rows.gz
+    cellmeta:
+      class: File
+      location: http://ftp.ebi.ac.uk/pub/databases/microarray/data/atlas/sc_experiments/E-MTAB-6077/E-MTAB-6077.cell_metadata.tsv 
+    barcodes:
+      class: File
+      location: http://ftp.ebi.ac.uk/pub/databases/microarray/data/atlas/sc_experiments/E-MTAB-6077/E-MTAB-6077.aggregated_counts.mtx_cols.gz
+    matrix:
+      class: File
+      location: http://ftp.ebi.ac.uk/pub/databases/microarray/data/atlas/sc_experiments/E-MTAB-6077/E-MTAB-6077.aggregated_counts.mtx.gz
--- a/workflows/single-cell-RNAseq/SingleCellExpressionAtlas-downstream/SingleCellExpressionAtlas-smartseq.ga
+++ b/workflows/single-cell-RNAseq/SingleCellExpressionAtlas-downstream/SingleCellExpressionAtlas-smartseq.ga
@@ -1,0 +1,1 @@
+../.ebi-gene-expression-group/scxa-workflows/w_smart-seq_clustering/scanpy_clustering_workflow.json


pcm32:
Workflow execution fails probably because our tools from the toolshed (user ebi-gxa) are not on the CVMFS setup. How can we add those there? Thanks!

mvdbeek:
Thanks @pcm32, those are really cool workflows!

> 1.- I would hope that the WorkflowHub.eu entry can be annotated to give credits to the person who wrote the workflow. I think that it will currently show as IWC, and while I of course value all the excellent feedback that the IWC community will provide, I still think that the main credit should go to the person writing the workflow (Jon Manning in this case).

You can set the author in the workflow file and it should appear in TRS interface ultimately (we have some work to do there, but it should be straightforward for dockstore at least).

> 2.- It would be great as well if, besides setting the author of the workflow to be passed to WorkflowHub.eu, additional organisations could be added (so that we could add, for instance, the EBI Gene Expression Team, next to IWC).

That is functionality workflowhub would need to expose, I'm not familiar enough to say if this is possible currently. This can be done on dockstore. You can definitely add multiple organizations in the workflow editor interface, and that will be the source of truth eventually.

> 3.- Our workflow setup currently allows certains steps to fail. For instance, when it comes to the cluster phase, I think that the process relies on some convergence criteria, and for some resolutions value a dataset might not converge if I remember well. Will the testing here consider the workflow failed if one of those steps failed? On our setup (https://github.com/ebi-gene-expression-group/galaxy-workflow-executor) we allow certain steps to fail while still considering the workflow to be successful.

No, I don't think that's something we'll allow in the intermediate future. We could think about re-running failed steps a couple of times though, would that help ? If we then still can't get the step to work reliably I would say that is a problem and not a workflow we should consider best-practice since it wouldn't work (without tuning the Galaxy instance) on any Galaxy instance.

> 4.- We keep our SC workflow versioned controlled on another repo, and I figured that the best way to keep this in-sync would be as a hidden git submodule. Open to suggestions if you would like to avoid this. I suspect it might break tests initially if the checkout behaviour is not doing submodules. Happy to fix this of course.

So the long term plan is that the infrastructure that we are developing for the iwc can be easily adopted by other groups (mostly through improving planemo-ci-action), and https://github.com/ebi-gene-expression-group/scxa-workflows looks like an excellent candidate for this approach. We could then still take up workflows that are published by ebi-gene-expression-group into a collection on dockstore.

I think the submodule approach would be a bit of a barrier both for contributions (although we wouldn't have to enforce this ...) and reviews, but I'm curious what other people in @galaxyproject/iwc think about this. It is certainly elegant.

mvdbeek:
Ah, and for installing the missing tools on cvmfs / usegalaxy.org, that would happen via a pull-request against https://github.com/galaxyproject/usegalaxy-tools. I will do this tomorrow, the plan is that we automate that.

pcm32:
>No, I don't think that's something we'll allow in the intermediate future. We could think about re-running failed steps a couple of times though, would that help ? If we then still can't get the step to work reliably I would say that is a problem and not a workflow we should consider best-practice since it wouldn't work (without tuning the Galaxy instance) on any Galaxy instance.

This is anyway on an intermediate step, which then gets a filtering of failed datasets on a collection, so the final outputs don't get affected anyway. But this is not due to a tool error, but more likely that the problem doesn't have a solution for that specific set of parameters: the workflow walks on a series of values for what is called the resolution of the clustering, and some combinations of datasets and resolution values won't have a solution, so some elements of that collection (related to the resolution parameter walking) will show up as errors, for some datasets.

pcm32:
Anyway, we'll see once the tools are on the CVMFS.

simleo:
> 1.- I would hope that the WorkflowHub.eu entry can be annotated to give credits to the person who wrote the workflow. I think that it will currently show as IWC, and while I of course value all the excellent feedback that the IWC community will provide, I still think that the main credit should go to the person writing the workflow (Jon Manning in this case).

If you set `"creator"` in the workflow file (see [this example](https://github.com/galaxyproject/iwc/blob/cfe57c80422b10d4a3ad22c5e1a33d0c9af9ae30/workflows/sars-cov-2-variant-calling/sars-cov-2-pe-illumina-artic-variant-calling/pe-artic-variation.ga#L4-L10)), it will show up in the [WorkflowHub page](https://workflowhub.eu/workflows/137) for the workflow:

Screenshot from 2021-10-06 12-49-21


https://github.com/galaxyproject/iwc/blob/cfe57c80422b10d4a3ad22c5e1a33d0c9af9ae30/workflows/sars-cov-2-variant-calling/sars-cov-2-pe-illumina-artic-variant-calling/pe-artic-variation.ga#L4

simleo:
> 2.- It would be great as well if, besides setting the author of the workflow to be passed to WorkflowHub.eu, additional organisations could be added (so that we could add, for instance, the EBI Gene Expression Team, next to IWC).

In [this example](https://github.com/galaxyproject/iwc/blob/cfe57c80422b10d4a3ad22c5e1a33d0c9af9ae30/workflows/data-fetching/parallel-accession-download/parallel-accession-download.ga#L4-L15), both `Person` and `Organization` are set in `creator`:

https://github.com/galaxyproject/iwc/blob/cfe57c80422b10d4a3ad22c5e1a33d0c9af9ae30/workflows/data-fetching/parallel-accession-download/parallel-accession-download.ga#L4-L15

But I think `Organization` is currently ignored by WorkflowHub. @fbacall can you confirm?

PR #66, opened 2021-10-13
Update SARS-CoV-2 Artic Illumina iVar workflow to use latest pangolin and nextclade

pvanheus:
This simply updates the tools in the workflow to the latest versions

### DIFF
--- a/workflows/sars-cov-2-variant-calling/sars-cov-2-pe-illumina-artic-ivar-analysis/pe-wgs-ivar-analysis.ga
+++ b/workflows/sars-cov-2-variant-calling/sars-cov-2-pe-illumina-artic-ivar-analysis/pe-wgs-ivar-analysis.ga
@@ -1136,5 +1136,5 @@
         "ARTIC",
         "iwc"
     ],
-    "uuid": "d784cd9f-b1a1-4a13-8400-5fc5ffa14dd0",
+    "uuid": "d784cd9f-b1a1-4a13-8400-5fc5ffa14dd0"
 }


pvanheus:
> A nice to have add-on could be to add actual nextclade and pangolin output test files, which will have to be updated with each release. Their diffs would directly show what's changed.

I don't think this is a good idea because these tools download databases as part of their runtime. The output of the tools would change while the tool version stays the same.

wm75:
> > A nice to have add-on could be to add actual nextclade and pangolin output test files, which will have to be updated with each release. Their diffs would directly show what's changed.
> 
> I don't think this is a good idea because these tools download databases as part of their runtime (see [here](https://github.com/galaxyproject/tools-iuc/blob/master/tools/nextclade/nextclade.xml#L13) and [here](https://github.com/galaxyproject/tools-iuc/blob/master/tools/pangolin/pangolin.xml#L12)). The output of the tools would change while the tool version stays the same.

Good point! Didn't think this through carefully, sorry.
As a compromise: would it hurt to set the header options of nextclade and pangolin to `true` to be able to check that columns in the tabular output stay the same, at least?

bwlang (approved):

bwlang:
@pvanheus , @wm75 
I'm proposing a couple of other minor tweaks here -  I've tested this workflow by importing and running on usegalaxy.eu
It complained a bit about tool versions but produced identical consensus sequences.

pvanheus:
> @pvanheus , @wm75 I'm proposing a couple of other minor tweaks here - I've tested this workflow by importing and running on usegalaxy.eu It complained a bit about tool versions but produced identical consensus sequences.

thanks... managing exact tool versions between Galaxy servers is quite a chore. In terms of getting "acceptable" results, this project - https://github.com/CDCgov/datasets-sars-cov-2 - is a start on benchmark datasets. We should have a discussion about appropriate use of these for validating this and other workflows.

PR #72, opened 2021-12-03
Fixes for computational-chemistry

simleo:
* Change workflow names in `.workflowhub.yml` to the `repo-name/dockstore-workflow-name` format
* Move test data inside workflow repo directories. Without this change, the Planemo test files in the target repositories (in `iwc-workflows`) would reference nonexistent files such as `"../test-data/lig.sdf"`


simonbray (approved):
Thank you @simleo

simonbray:
Not sure if the version numbers need to be bumped, at least for `fragment-based-docking-scoring.ga`, which is the only one that the CI successfully deployed: https://github.com/iwc-workflows/fragment-based-docking-scoring

simleo:
> Not sure if the version numbers need to be bumped

Yes, thanks for reminding me. I've changed the micro version since workflows did not change.

PR #74, opened 2021-12-13
Workflow-specific CI workflows

### LINKED ISSUE #71: Workflow-specific CI runs
For clarity, in the discussion that follows, I will use "workflow" for a scientific workflow and "gh-workflow" for a GitHub Actions workflow.

The fact that the same gh-workflow is run to test all workflows is a problem for monitoring. Since all RO-Crates point to https://api.github.com/repos/galaxyproject/iwc/actions/workflows/workflow_test.yml, the [LifeMonitor dashboard](https://app.lifemonitor.eu/dashboard) ends up showing the same health status and test build history for all workflows, which is not particularly informative. For instance, if the test for `parallel-accession-download/main` fails, the build will be marked as a failure for all workflows, even if all the others are doing fine.

For this reason, it would be useful to run a separate gh-workflow for each workflow. One solution would be to have all these separate gh-workflows in this repository. However, they would have to be added manually by the author of each new workflow. Also, as the number of workflows in the repo increases, its GitHub Actions history could become pretty cluttered. Adding them to each individual repository in iwc-workflows, however, should work. Here are some considerations on the individual gh-workflows:

* They can be much simpler than the main one in iwc, since they only have to run the tests (no deployment, etc.).
* How to author them? They can be generated before repo deployment in the main gh-workflow, much like RO-Crate metadata files are generated now.

* When should they be run? Not at each commit: that would be useless, since updates to the individual repositories are only made after tests run successfully in the main gh-workflow. Instead, they should be run periodically (e.g., with a `cron: '0 3 * * *'`) to check for failures due to changes in dependencies, external references, etc.

simleo:
Resolves #71.

The script that generates RO-Crate metadata files now also places a GitHub workflow that tests the scientific workflow under `.github/workflows`, allowing each deployed repository to have its own GitHub Actions history reflecting the status of its specific workflow. The GitHub workflow is set to run on a daily schedule and it's also configured to enable manual triggering; I've tested it on a fork of `iwc-workflows/sars-cov-2-pe-illumina-artic-variant-calling`, see https://github.com/simleo/sars-cov-2-pe-illumina-artic-variant-calling/actions.

### DIFF
--- a/workflows/gen_crates.py
+++ b/workflows/gen_crates.py
@@ -54,8 +54,8 @@ try:
 except ImportError:
     from yaml import Loader
 
-OWNER = "galaxyproject"
-REPO = "iwc"
+THIS_DIR = Path(__file__).absolute().parent
+
 GH_WORKFLOW = "workflow_test.yml"
 TARGET_OWNER = "iwc-workflows"
 GH_API_URL = "https://api.github.com"
@@ -75,6 +75,7 @@ DOCKSTORE_CFG_NAME = ".dockstore.yml"
 DOCKSTORE_CFG_VERSION = "1.2"
 LM_URL = "https://api.lifemonitor.eu"
 LM_API_KEY = os.getenv("LM_API_KEY")
+CI_WORKFLOW = THIS_DIR / "wftest.yml"
 
 
 class HubClient:
@@ -229,7 +230,7 @@ def get_workflow_name(repo_dir):
     return f"{repo_dir.name}/{wf_entry['name']}"
 
 
-def make_crate(crate_dir, target_owner, resource, planemo_version):
+def make_crate(crate_dir, target_owner, planemo_version):
     wf_id = get_wf_id(crate_dir)
     planemo_id, planemo_source = get_planemo_id(crate_dir, wf_id)
     crate = ROCrate(gen_preview=False)
@@ -254,6 +255,7 @@ def make_crate(crate_dir, target_owner, resource, planemo_version):
     if readme_source.is_file():
         crate.add_file(readme_source, "README.md")
     suite = crate.add_test_suite(identifier="#test1")
+    resource = f"repos/{target_owner}/{crate_dir.name}/actions/workflows/{CI_WORKFLOW.name}"
     crate.add_test_instance(suite, GH_API_URL, resource=resource,
                             service="github", identifier="test1_1")
     crate.add_test_definition(suite, source=planemo_source,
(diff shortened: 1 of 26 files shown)


workflows/wftest.yml
```diff
@@ -0,0 +1,23 @@
+name: Periodic workflow test
```
mvdbeek, line 1: I think this looks good in principle, but we should run another workflow (ideally the test portion of the workflow in the iwc repo) here via [reusable workflows](https://docs.github.com/en/actions/learn-github-actions/reusing-workflows#creating-a-reusable-workflow). I don't think we'll want to keep separate copies that we have to update with each new version of the parent workflow. The workflow also lacks all of the caching steps, I think we want to keep this, which should be doable with the reusable workflow setup. That's a bit more work but I think it'll pay off pretty soon.

workflows/gen_crates.py
```diff
@@ -343,13 +345,15 @@ def main(args):
...
     if args.upload_to_hub or args.upload_to_lm:
         client = HubClient(base_url=args.hub_url)
-    resource = f"repos/{args.owner}/{args.repo}/actions/workflows/{args.workflow}"
     for repo in find_repos(args.root, exclude=args.exclude):
         print(f"processing {repo}")
         if args.no_overwrite and (repo / "ro-crate-metadata.json").is_file():
             print("  crate exists, not overwriting")
         else:
-            make_crate(repo, args.target_owner, resource, args.planemo_version)
+            target_ci_workflow = repo / ".github" / "workflows" / CI_WORKFLOW.name
```
mvdbeek, line 353: Ideally we'd upstream this into planemo, but that's fine for now.

workflows/wftest.yml
simleo: Here's the [reusable workflow](https://github.com/galaxyproject/iwc/blob/f7a3e1e6c49bc1c6afd6a056318d0154c05db715/.github/workflows/test_workflows.yml) for galaxy workflow testing. For now I've tested it with a [caller](https://github.com/simleo/sars-cov-2-se-illumina-wgs-variant-calling/blob/dda19bbb989be432909d6540ae2887c0aa47bc1d/.github/workflows/wftest.yml) written for individual `iwc-workflows` repos, in https://github.com/simleo/sars-cov-2-se-illumina-wgs-variant-calling. It seems to be working fine. I still have to check its integration in the main `iwc` workflow.

mvdbeek: That looks amazing, thanks so much @simleo!

simleo:
806de2d changes the per-repo CI workflow so that it calls the [reusable workflows](https://github.com/galaxyproject/iwc/pull/75) for setup and test. It's been tested [here](https://github.com/simleo/sars-cov-2-se-illumina-wgs-variant-calling/tree/e697a0da1037ecd62bd4db04e71bf3fb11d81b87).

mvdbeek (approved):
Very cool, thank you so much @simleo!

Issue #58, opened 2021-09-13
Workflow autoupdates


simonbray:
Parallel to https://github.com/galaxyproject/tools-iuc/issues/3533, it would be great to implement automatic workflow updates.

I wrote a planemo PR which should achieve that here: https://github.com/galaxyproject/planemo/pull/1151. Basically it does exactly the same thing as a user who 1) manually uploads a workflow to a Galaxy server with the latest versions of all tools installed, 2) presses the `Upgrade All Workflow Steps` button, and 3) redownloads the updated workflow.

Unfortunately the resulting diffs are currently a horrible mess. Here is an example: https://github.com/galaxyproject/iwc/compare/main...simonbray:example-autoupdate

Not sure if anyone has any ideas about how or whether this can be improved. My other idea was to implement a much simpler approach of just iterating over the workflow steps and checking the workflow ID against the toolshed(s) for any newer versions. This should give much cleaner diffs and should also be a lot faster (no need to start up a Galaxy server and install all the necessary tools). But running the autoupgrade through a Galaxy server should be more sophisticated in correcting for e.g. added or deleted tool parameters.

mvdbeek:
We'll need to gxformat2 with completely labelled steps, this should be much cleaner.

> But running the autoupgrade through a Galaxy server should be more sophisticated in correcting for e.g. added or deleted tool parameters.

Yes, I think that is necessary

simonbray:
> We'll need to gxformat2 with completely labelled steps, this should be much cleaner.

So we should put this on hold until the IWC is operating using gxformat2 workflows?

mvdbeek:
No, this is great. We're also pretty much ready for gxformat2.

PR #79, opened 2022-01-18
Updating workflows/computational-chemistry/protein-ligand-complex-parameterization

planemo-autoupdate:
Hello! This is an automated update of the following workflow: **workflows/computational-chemistry/protein-ligand-complex-parameterization**. I created this PR because I think one or more of the component tools are out of date, i.e. there is a newer version available on the ToolShed.


simonbray:
An initial implementation of #58 - as discussed in that issue, I think we would need to switch to gxformat2 for this really to make sense.

workflows/computational-chemistry/protein-ligand-complex-parameterization/protein-ligand-complex-parameterization.ga
```diff
@@ -77,6 +76,253 @@
             "uuid": "277627e0-eba8-4de5-855b-63711c0db691",
             "workflow_outputs": []
         },
+        "10": {
```
mvdbeek, line 79: @simonbray This does look suspiciously like the order is messed up because the sorting wasn't numeric ... any chance you could see if that could be fixed ? That might be more feasible in the short term than switching to gxformat2

simonbray: Yes, thanks for spotting that!!

workflows/computational-chemistry/protein-ligand-complex-parameterization/protein-ligand-complex-parameterization.ga
```diff
@@ -11,7 +11,6 @@
     "format-version": "0.1",
     "license": "MIT",
     "name": "Create GRO and TOP complex files",
-    "release": "0.1.2",
```
simonbray, line 14: should not be removed

PR #80, opened 2022-01-18
Updating workflows/computational-chemistry/protein-ligand-complex-parameterization

planemo-autoupdate:
Hello! This is an automated update of the following workflow: **workflows/computational-chemistry/protein-ligand-complex-parameterization**. I created this PR because I think one or more of the component tools are out of date, i.e. there is a newer version available on the ToolShed.

### DIFF
--- a/workflows/computational-chemistry/protein-ligand-complex-parameterization/README.md
+++ b/workflows/computational-chemistry/protein-ligand-complex-parameterization/README.md
@@ -5,4 +5,5 @@ dynamics simulation with GROMACS.
 
 This is a simple workflow intended for use as a subworkflow in more complex
 MD workflows. It is used as a subworkflow by the GROMACS MMGBSA and dcTMD
-workflows. 
+workflows.
+


workflows/computational-chemistry/protein-ligand-complex-parameterization/protein-ligand-complex-parameterization.ga
```diff
@@ -266,7 +266,7 @@
             },
             "tool_id": "toolshed.g2.bx.psu.edu/repos/bgruening/ctb_rdkit_descriptors/ctb_rdkit_descriptors/2020.03.4+galaxy1",
             "tool_shed_repository": {
-                "changeset_revision": "a1c53f0533b0",
+                "changeset_revision": "0993ac4f4a23",
```
mvdbeek, line 269: Wow, that looks very cool. But it seems there were no version bumps, is that correct ?

simonbray: Yes, there are none (the updated `changeset_revision`s are due to other tools in the repos being updated).

So this needs a bit more work.

simonbray: This should be fixed now, I experimented a bit more on my fork: https://github.com/simonbray/iwc/pulls.

PR #83, opened 2022-02-02
Recalculate AFs before consensus building

wm75:
Version 0.3 of the Consensus from Variation workflow with significantly improved classification of variants into consensus and ambiguous variants.


mvdbeek (approved):
Nice, thank you @wm75!

PR #95, opened 2022-02-11
Fix AF and DP_ALT handling in reporting workflow

wm75:


wm75:
In the by-variants report both the list of all samples with a given variant and the corresponding list of AFs are dependent on the order of the input samples in the collection, which I had built differently in my own testing. Should be fixed now.

wm75:
but there seems to be some general issue with running tests currently

wm75:
Seems to have been a temporary issue that also impacted the weekly tests of all WFs.
Tests are passing now again :)

mvdbeek:
Yeah, looks like the cvmfs apt repository was down.

wm75:
Don't want to merge my own PR, but I think it's ready :)

PR #97, opened 2022-03-17
Script and GH action for IWC workflow installation on usegalaxy.*

simonbray:
This would be nicer if https://github.com/galaxyproject/galaxy/pull/13376 was available, but that needs to wait until 22.05. In the meantime the script includes a workaround where a dataset in the IWC account is tagged with all TRS IDs of the already installed workflows. When the script is run (weekly) it checks Dockstore for IWC workflows, compares with the list of already-installed TRS IDs, installs and publishes any missing ones, and updates the list of tags.

### DIFF
--- a/scripts/usegalaxy_star_upload.py
+++ b/scripts/usegalaxy_star_upload.py
@@ -1,0 +1,90 @@
+import os
+import logging
+
+import requests
+
+from bioblend import galaxy
+
+DOCKSTORE_API_URL = "https://dockstore.org/api/ga4gh/trs/v2/tools"
+DOCKSTORE_API_PARAMS = {"organization": "iwc-workflows"}
+
+GALAXIES_TO_UPDATE_CREDENTIALS = [
+    {"url": "https://usegalaxy.org", "key": os.getenv("IWC_KEY_USEGALAXY_ORG")},
+    {"url": "https://usegalaxy.eu", "key": os.getenv("IWC_KEY_USEGALAXY_EU")},
+    {"url": "https://usegalaxy.org.au", "key": os.getenv("IWC_KEY_USEGALAXY_ORG_AU")},
+]
+
+logging.getLogger().setLevel(logging.INFO)
+
+# get all IWC workflows from Dockstore
+dockstore_wfs = requests.get(DOCKSTORE_API_URL, params=DOCKSTORE_API_PARAMS).json()
+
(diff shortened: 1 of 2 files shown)


scripts/usegalaxy_star_upload.py
```diff
@@ -0,0 +1,90 @@
...
+        continue
+    gi = galaxy.GalaxyInstance(**galaxy_credentials)
+
+    # get all current workflows on the Galaxy server
+    galaxy_wfs = gi.workflows.get_workflows()
+    # a hack: we use a dataset which is tagged with the TRS ids of all already published workflows.
+    # before installing new workflows, we check if the TRS id is in this list and if not we install.
+    # when https://github.com/galaxyproject/galaxy/pull/13541 is available on all servers the workflows
+    # will be annotated directly with the TRS ids and we can remove this.
+    tagged_dataset = gi.datasets.get_datasets(limit=1, order="create_time-asc")[0]
```
mvdbeek, line 35: You can also check the workflow uuid, that seems like a much more direct check

simonbray: It doesn't seem to be constant, otherwise that would be great.

simonbray: Sorry, that's not correct. It is constant.

mvdbeek: It is constant for the workflow you upload.

simonbray: But another thought: what if an IWC PR gets merged without updating the uuid? Then the uuid will not be unique and the new version won't get installed.

mvdbeek: 😆 sorry, github UI didn't update in time

simonbray: > But another thought: what if an IWC PR gets merged without updating the uuid? Then the uuid will not be unique and the new version won't get installed.

I found a concrete example; releases 0.1.1 and 0.1.2 of the gromacs-dctmd workflow have the same uuid:
* https://github.com/iwc-workflows/gromacs-dctmd/blob/v0.1.1/gromacs-dctmd.ga#L2487
* https://github.com/iwc-workflows/gromacs-dctmd/blob/v0.1.2/gromacs-dctmd.ga#L2487

PR #112, opened 2022-07-01
adding galaxy workflow of VGP_decontamination to iwc 

Nadolina:


bgruening:
Cool thanks @Nadolina

mvdbeek:
Thanks for the contribution @Nadolina! Are you up following [these steps](https://github.com/galaxyproject/iwc/blob/main/workflows/README.md#adding-workflows) ? If not, could you provide us with a history where you've run the workflow on some small test data ?

Nadolina:
@mvdbeek, thanks for your reply! I would prefer to provide a history - this one uses a small fasta file. Here is the link: https://assembly.usegalaxy.eu/u/nadolina/h/testdatadecon

PR #127, opened 2022-10-18
ChIP SR

lldelisle:
New workflow ChIP for Single Read. It is the corresponding ChIP PE but for single read.


lldelisle:
Ideally, I would have liked to propose the user to choose between 'model' and 'nomodel' for macs2 but it seems that it cannot be a workflow parameter as it is a conditional.

mvdbeek:
Hmm, this might be another case where we need conditionals.

lldelisle:
You mean where we need conditional steps, right?

mvdbeek:
Yes!

lldelisle:
Meanwhile we could do subworkflows but I don't know if it worths.

mvdbeek:
So far probably not, but I think we could ultimately share the multiqc/big/macs2 summary steps with the PE workflow via gxformat2 references. Can you also change the name to lowercase (chip-seq-sr or chipseq-sr)?

lldelisle:
Even if ChIPseq_PE is in upper case?
Or should I modify both?

mvdbeek:
Yeah, I thought about that too late, but if you can change both that would be great. I can rename the PE one in github/dockstore/workflowhub

lldelisle:
Should I write something in the changelog of PE?
Can we do something as 'mark as deprecated' in both repositories?

lldelisle:
:tada:

Issue #120, opened 2022-09-30
Mention `planemo test --galaxy_url https://usegalaxy.org/ --galaxy_user_key <your_api_key> <workflow.ga>` for trouble shooting


mvdbeek:

bernt-matthias:
Fixed in https://github.com/galaxyproject/iwc/pull/129

PR #129, opened 2022-10-19
improve README in workflows

lldelisle:
Following #120 
Inputs are welcome

### DIFF
--- a/workflows/README.md
+++ b/workflows/README.md
@@ -2,6 +2,8 @@
 
 This directory contains all IWC workflows.
 
+## Structure of the directory
+
 The structure is as follows:
 
 * top-level directories represent categories, e.g., `sars-cov-2-variant-calling`;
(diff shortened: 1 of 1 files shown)


workflows/README.md
```diff
@@ -16,13 +18,21 @@ The structure is as follows:
...
 
 It is recommended to check and update tools with Planemo before import :
+
 ```bash
-planemo autoupdate my_workflow.ga
+planemo autoupdate <workflow.ga>
 ```
-Then test the updated workflow from inside usegalaxy.* server. Using GTN dataset material is a good start !
+
+Then, it is recommanded to upload this updated workflow inside usegalaxy.org server as some updates may break connections (and some tools may not have the last update).
```
mvdbeek, line 35:
```suggestion
Then, it is recommended to upload this updated workflow to the usegalaxy.org server as some updates may break connections (and some tools may not yet be installed on usegalaxy.org).
```

mvdbeek (approved):
Thank you!

PR #130, opened 2022-10-21
Rnaseq PE

lldelisle:
Here is the sister of RNA-seq SR


mvdbeek (approved):
Very nice, thank you @lldelisle! I'm working on conditionals now, I hope we can use this to maybe merge the sr/pe workflows with the next Galaxy release.

PR #131, opened 2022-10-21
Fix and simplify SARS-CoV-2 consensus WF

wm75:


wm75:
Success on first pass would have been to good to be true I guess, but with the release now bumbed things should be good for merging :)

mvdbeek (approved):
Nice, thanks for the update @wm75!

PR #132, opened 2022-10-25
[WIP] Hi-C workflow

lldelisle:
I tried to use subworkflows.
I would like to have 2 'main' workflows: Hi-C and tiled-region capture Hi-C the only difference is a filtering step... But waiting for the conditional step I think I need 2 workflows.


mvdbeek:
Hmm, that's interesting, I'll take a look, hopefully this is just a minor issue.

lldelisle:
Also, for the moment these tools cannot be installed in usegalaxy.org as `planemo monitor` is failing: https://github.com/galaxyproject/planemo-monitor/issues/41

lldelisle:
### Here is the situation:
- I have 2 main workflows + 2 subworkflows
- I put tests for the 2 main workflows and 1 subworkflow, for the other one I could not put tests see below.
- All tests are passing on my instance and failing in the CI because the tools are not yet installed on usegalaxy.org but this would need some time as the cooler tools are on the tools-lldelisle so have only been reviewed by myself...

### Why I cannot put a test for the subworkflow, which is the same reason as why my tool cooler_cload_tabix has not test:
The input is a tabix format, this format is not available at upload, therefore in the last version of the wrapper I also accept bgzip format which is available at upload and bgzip to tabix is just 1 line of code.
However, I discovered that when I upload a bgzip in galaxy (version 22.05) the file is decompressed but still keep the 'dtype' bgzip. So my test is failing with a `the compression of ... is not BGZF`.
Then I found a trick (which I wrote in this README), if you download your tabix file as 'interval' then you set the metadata so chromosome column is 3 and start and end columns are 4, you can run the workflow and it works (I tested it on my instance). Indeed there is a interval_to_bgzip converter which uses the metadata. However, while it works manually, I don't manage to set the metadata correctly to be set at upload (see https://github.com/galaxyproject/galaxy/pull/14869). This is the case in planemo, usegalaxy.org (version 22.05) and on my instance (version 21.01).

### What do we do?
- Accept a sub-workflow without test and accept in usegalaxy.org a tool that does not have test...
- We fix the bgzip upload
- We fix the interval upload with metadata

### What can I do?
So I don't know what to do, my knowledge in the galaxy code is too low to understand why when I upload a bgzip it is decompressed but if you need help to write a sniffer for bgzip, I think I may be able do it using http://samtools.github.io/hts-specs/SAMv1.pdf section 4 and https://github.com/galaxyproject/galaxy/blob/d5295de026611fea7fa965fccb6e9c388836b315/lib/galaxy/datatypes/tabular.py#L1041-L1047. I also don't know why the metadata is not used at upload...

Sorry...

mvdbeek:
> What do we do?

Isn't there the 4th option of enabling the upload of tabix files (and making sure they stay compressed) ?

lldelisle:
Indeed this could be an option, in theory, tabix files always comes with a tbi (index) but I think we could do call tabix without it...

mvdbeek:
I think that planemo call should work, the error looks like a problem with the SSL certificates. If you're doing CI testing you could set galaxy_branch and galaxy_source to rule out a local problem (in https://github.com/lldelisle/tools-lldelisle/blob/master/.github/workflows/ci.yaml#L9-L10 / https://github.com/lldelisle/tools-lldelisle/blob/master/.github/workflows/pr.yaml#L4-L5)

simleo:
WorkflowHub / Workflow RO-Crate does not support more than one main workflow. The `gen_crates.py` script does not break in this case: it picks one of the workflows as the main workflow (currently, the first in the `os.scandir` list). Also, the workflow name is set to `REPO_NAME/DOCKSTORE_NAME` = `hic-hicup-cooler/main`, again using the first record encountered while reading `.dockstore.yaml`. In this case, all the workflows in `.dockstore.yaml` have the same name: it looks like [this is not supported by Dockstore](https://docs.dockstore.org/en/stable/assets/templates/workflows/workflows.html#full-template-with-explanation-of-all-available-fields). Actually, in its current version, `gen_crates.py` is not prepared for multiple workflows even if only one of them is the main one and the others are subworkflows: I am working on this, but we need a way to let it know which one is the main workflow, so it does not just arbitrarily pick one.

To summarize, here is the list of changes that should make things compatible with all services:

* Split the repository into separate repositories, one for each main workflow
* In each repository, place the main workflow together with the subworkflows it needs and the corresponding Planemo test config files. In `.dockstore.yml` use different, unique names for all workflows, and list the main workflow as the **first entry**

mvdbeek:
I would not want to split related workflows into multiple repositories, and that is not necessary for dockstore, you simply
use a different name. `main` is just what we use when there is a single workflow. Can this be supported by workflowhub ?

workflows/epigenetics/hic-hicup-cooler/README.md
```diff
@@ -0,0 +1,61 @@
...
+- The output BAM file is converted to juicer format: `<readname> <str1> <chr1> <pos1> <frag1> <str2> <chr2> <pos2> <frag2> <mapq1> <mapq2>` where str = strand (0 for forward, anything else for reverse) and pos is the 5' end.
+- The pairs are filtered for MAPQ if specified.
+- For the region capture Hi-C workflow the pairs are filtered for both mates in the captured region.
+- The filtered pairs are sorted and indexed with cooler_csort.
+- The pairs are loaded into a matrix of the given resolution and balanced with cooler.
+
+## Subworkflows
+
+There are 2 subworkflows: `hic_tabix_to_cool_cooler` and `hic_fastq_to_pairs_hicup.ga`.
+
```
mvdbeek, line 39: We should be able to launch subworkflows through the Galaxy interface (maybe with a dropdown icon or something like this), perhaps we don't need to include them as installable workflows ?

workflows/epigenetics/hic-hicup-cooler/.dockstore.yml
```diff
@@ -0,0 +1,20 @@
...
+- name: main
+  primaryDescriptorPath: /hic_fastq_to_cool_hicup_cooler.ga
+  subclass: Galaxy
+  testParameterFiles:
+  - /hic_fastq_to_cool_hicup_cooler-tests.yml
+- name: main
+  primaryDescriptorPath: /hic_fastq_to_pairs_hicup.ga
+  subclass: Galaxy
+  testParameterFiles:
+  - /hic_fastq_to_pairs_hicup-tests.yml
```
mvdbeek, line 20:
```suggestion
- name: hic-tabix-cool-cooler
  primaryDescriptorPath: /hic_tabix_to_cool_cooler.ga
  subclass: Galaxy
- name: chic-fastq-to-cool-hicup-cooler
  primaryDescriptorPath: /chic_fastq_to_cool_hicup_cooler.ga
  subclass: Galaxy
  testParameterFiles:
  - /chic_fastq_to_cool_hicup_cooler-tests.yml
- name: hic-fastq-to-cool-hicup-cooler.
  primaryDescriptorPath: /hic_fastq_to_cool_hicup_cooler.ga
  subclass: Galaxy
  testParameterFiles:
  - /hic_fastq_to_cool_hicup_cooler-tests.yml
- name: hic-fastq-to-pairs-hicup
  primaryDescriptorPath: /hic_fastq_to_pairs_hicup.ga
  subclass: Galaxy
  testParameterFiles:
  - /hic_fastq_to_pairs_hicup-tests.yml
```
main is what we use if there's a single workflow for the whole repo

lldelisle: Should we add the version in the name?

workflows/epigenetics/hic-hicup-cooler/README.md
lldelisle: You mean we don't need to make them appear from dockstore or workflowhub?

workflows/epigenetics/hic-hicup-cooler/.dockstore.yml
mvdbeek: No, the version is tracked separately

workflows/epigenetics/hic-hicup-cooler/README.md
mvdbeek: I think if we change the Galaxy interface to be able to select subworkflows we might not need it to be published separately ... but for now that feature doesn't exist, so let's keep them all. Sorry for the noise 😅

mvdbeek: And still I think even if the feature exists it would be nice if one day you can write `planemo run #workflow/github.com/iwc-workflows/hic-hicup-cooler/hic_tabix_to_cool_cooler `, and for that we need to list them all :)

simleo:
> I would not want to split related workflows into multiple repositories, and that is not necessary for dockstore, you simply use a different name. `main` is just what we use when there is a single workflow. Can this be supported by workflowhub ?

At the moment it's one Workflow RO-Crate / one WorkflowHub entry / one LifeMonitor entry, which I think is a strong assumption by both systems.

On the Workflow RO-Crate side, we'd need to allow an array of values for `mainEntity`. Syntactically this is allowed by JSON-LD, though I'm not sure it would make sense semantically. @stain what do you think?

Then WorkflowHub and LifeMonitor would have to be prepared to generate multiple entries when encountering such crates. However, there would be multiple workflows registered that have the same crate. I think this is the main difference with Dockstore: while `.dockstore.yml` is a configuration file to generate registry entries, the RO-Crate metadata file describes the workflow and "stays with" the workflow, whose RO-Crate can be downloaded again afterwards, like LifeMonitor does when registering a workflow from WorkflowHub. With a multi-workflow crate, what should LifeMonitor do? Update all corresponding entries, or try to filter for metadata corresponding to the WorkflowHub entry? And what if the registration is not from WorkflowHub, but directly from the RO-Crate? @fbacall @kikkomep @ilveroluca thoughts?

Last but not least, how should `gen_crates.py` distinguish main workflows from subworkflows then? Use `publish: false` for subworkflows in `.dockstore.yml`?

workflows/epigenetics/hic-hicup-cooler/README.md
lldelisle: So I mark this as resolved.

mvdbeek:
What would you do if there's more than one workflow in a repository, i.e. a monorepo ? That seems like quite the common case ?


> At the moment it's one Workflow RO-Crate / one WorkflowHub entry / one LifeMonitor entry, which I think is a strong assumption by both systems.

And why would that have to change ? Can you not just create multiple entries from one repo ?

> Last but not least, how should `gen_crates.py` distinguish main workflows from subworkflows then? Use `publish: false` for subworkflows in `.dockstore.yml`?

There's no difference, a subworkflow is a workflow.

mvdbeek:
Not from the IWC, but https://github.com/astrovsky01/ASHG_Galaxy_Workflows is the source for https://dockstore.org/workflows/github.com/astrovsky01/ASHG_Galaxy_Workflows/ASHG_Plink_filtering_part_2_and_stratification:main?tab=info, https://dockstore.org/workflows/github.com/astrovsky01/ASHG_Galaxy_Workflows/ASHG_Plink_association_testing_and_plotting:main?tab=info, https://dockstore.org/workflows/github.com/astrovsky01/ASHG_Galaxy_Workflows/ASHG_Plink_sex_and_phenotype_assignment:main?tab=info, https://dockstore.org/workflows/github.com/astrovsky01/ASHG_Galaxy_Workflows/ASHG_Plink_filtering_part_1:main?tab=info
Those are registered manually though

lldelisle:
I found one:
https://github.com/galaxyproject/training-material/blob/main/.dockstore.yml
At the end in the dockstore we don't see they were on the same file right?

simleo:
> And why would that have to change ? Can you not just create multiple entries from one repo ?

An RO-Crate is essentially a directory with a metadata file at the top level that describes its contents. The metadata file has the fixed name `ro-crate-metadata.json`, so you can't add one separate metadata file for each workflow: you need to make a separate copy of the whole repository for each workflow and add a different `ro-crate-metadata.json` to each. This breaks the current one-to-one relationship between workflow repositories and RO-Crates, so it would be impossible to just add the metadata file to the repository before its publication in iwc-workflows like we're doing now. The crates would have to be generated on the fly when the repository is processed for registration to WorkflowHub. I guess that should be possible with the LifeMonitor GitHub app (which we are now testing), by installing it for the iwc-workflows organization. The only way to keep `ro-crate-metadata.json` hosted on GitHub would be to deploy separate iwc-workflows repositories for the same iwc repository in the multi-workflow case, but I don't think that's what you want.

fbacall:
@simleo 

Multiple `mainEntity`s in an RO-Crate doesn't make sense to me. Its purpose is to indicate the single primary entity if it's ambiguous. 

Creating 1 RO-Crate and registering 1 WorkflowHub entry per workflow (as listed in `.dockstore.yml`) makes the most sense to me. 

Currently WorkflowHub expects to find `ro-crate-metadata.json` at the root of the repository, which is problematic for the case where there are multiple workflows. We've had plans to allow users to provide a "root" path within the repository, which WorkflowHub could attempt to discover `ro-crate-metadata.json`, but nothing is implemented yet.

I was under the impression though, that the https://github.com/iwc-workflows org was created to publish separate workflow repos from this monorepo.

mvdbeek:
> I was under the impression though, that the https://github.com/iwc-workflows org was created to publish separate workflow repos from this monorepo.

it is, but it just doesn't make sense to split worklfows that belong together and that should be released as one version because they share logic.



> Creating 1 RO-Crate and registering 1 WorkflowHub entry per workflow (as listed in `.dockstore.yml`) makes the most sense to me.

Maybe that wasn't clear in my previous question, what's the reason you can't keep doing that, just create as many crates as needed.

simleo:
Discussed this morning with the LifeMonitor team. At the moment, the GitHub app is quite reliant on the assumption of a GitHub-hosted Workflow RO-Crate. We need some time to assess the impact of the changes that a multi-workflow repository entails.

stain:
If this repository is effectively a collection and does not have a "main" workflow, then it can't be registered as a _Workflow_ in WorkflowHub.  Perhaps it could be a [Collection](https://workflowhub.eu/collections)? 

We could have one `ro-crate-metadata.json` per workflow in some subfolders, each with a `mainEntity` for their workflow -- with an aggregating `ro-crate-metadata.json` at the root -- nested crates.   WorkflowHub does not support these at the moment, so this would be a new pattern.

The alternative is to have all the workflows in a shared `ro-crate-metadata.json` but no dedicated `mainEntity` -- WorkflowHub would still not know which one to pick so would need itself to generate derived crates, one per workflow -- perhaps it just is the same crate but modified `mainEntity`?

simleo:
Another problem with multi-workflow repositories like this one is that -- regardless of how their submission to WorkflowHub is handled -- LifeMonitor would not be able to separately track the test suites for the different workflows. The Python script could generate a separate CI workflow file for each workflow (rather than the current single `wftest.yml`), but planemo-ci-action does not allow to specify a single Planemo test definition, e.g. `repository-list: './chic_fastq_to_cool_hicup_cooler.ga'` (see [this CI run](https://github.com/simleo/hic-hicup-cooler-ci-test/actions/runs/3566560468/jobs/5993181751))

mvdbeek:
> Another problem with multi-workflow repositories like this one is that -- regardless of how their submission to WorkflowHub is handled -- LifeMonitor would not be able to separately track the test suites for the different workflows.

I get this is a problem for you, but I insist that it is a common thing to have more than one workflow in a repository.

> planemo-ci-action does not allow to specify a single Planemo test definition, e.g. `repository-list: './chic_fastq_to_cool_hicup_cooler.ga'` (see [this CI run](https://github.com/simleo/hic-hicup-cooler-ci-test/actions/runs/3566560468/jobs/5993181751))

Can you explain why you don't test all of the workflows at once ? You can split them into chunks if you like, just like what we do for PRs.

simleo:
> Can you explain why you don't test all of the workflows at once ? You can split them into chunks if you like, just like what we do for PRs.

Workflows can certainly be tested all at once, but then they would have the same CI build history, and you'd see a failure notification for all of them even when only one is failing.

simleo:
> That seems fine ? The aggregate state should be failing in that case ?

I guess it's a matter of what the developer / maintainer expects. However, knowing at a glance which workflow is failing and being able to read only the corresponding log can save time.

lldelisle:
Hi,
I am now blocked by https://github.com/galaxyproject/planemo/issues/1345

lldelisle:
Hi,
This PR is ready for final review.
I propose that for the moment these 4 workflows only go to the dockstore and not to the workflow hub, or I create one subdirectory per workflow and I create symlink for each...

simleo:
There's a WorkflowHub meeting tomorrow to discuss the handling of multi-workflow repositories. For now, I've [changed the script that generates the RO-Crate](https://github.com/galaxyproject/iwc/pull/170) to add all workflows listed in the Dockstore file. In its current version, WorkflowHub will still recognize only the one listed as the `mainEntity`, but with this change the information on the other ones will be there to be leveraged in future versions. #170 should be merged before this PR

lldelisle:
Thank you @simleo we indeed plan to wait for the meeting of tomorrow before merging this PR.

lldelisle:
@simleo should I add a .workflowhub.yml with a chosen mainEntity? Or should I leave it like this?

lldelisle:
Seems ready... Who wants to click?

mvdbeek (approved):
Fantastic work as always @lldelisle!

mvdbeek:
Also thanks to the workflowhub team!

PR #142, opened 2022-11-02
add missing CHANGELOG

lldelisle:
autoupdate failed because one workflow does not have changelog.
We should have in the CI something like:
```
for f in $(find . -name "*.ga"); do
  dir=$(dirname $f)
  if [ ! -e "$dir/README.md" ]; then
    echo "$dir has no README"
  fi
  if [ ! -e "$dir/CHANGELOG.md" ]; then
    echo "$dir has no CHANGELOG"
  fi
  if [ ! -e "$dir/.dockstore.yml" ]; then
    echo "$dir has no dockstore"
  fi
  if [ ! -e "$dir/.workflowhub.yml" ]; then
    echo "$dir has no workflowhub"
  fi
done
```

### DIFF
--- a/workflows/VGP-assembly-v2/VGP-meryldb-creation-trio/CHANGELOG.md
+++ b/workflows/VGP-assembly-v2/VGP-meryldb-creation-trio/CHANGELOG.md
@@ -1,0 +1,12 @@
+# Changelog
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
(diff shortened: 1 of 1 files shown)


lldelisle:
@gallardoalba Can you confirm the changelog is correct (I just copied it from the other VGP_v2 workflow)?

gallardoalba:
> @gallardoalba Can you confirm the changelog is correct (I just copied it from the other VGP_v2 workflow)?

Indeed one of the VGP workflows doesn't include a CHANGELOG file https://github.com/galaxyproject/iwc/tree/main/workflows/VGP-assembly-v2/VGP-meryldb-creation-trio

gallardoalba (approved):

PR #148, opened 2022-11-02
Updating workflows/transcriptomics/rnaseq-pe from 0.1 to 0.2

planemo-autoupdate:
Hello! This is an automated update of the following workflow: **workflows/transcriptomics/rnaseq-pe**. I created this PR because I think one or more of the component tools are out of date, i.e. there is a newer version available on the ToolShed.

* `toolshed.g2.bx.psu.edu/repos/iuc/rgrnastar/rna_star/2.7.8a+galaxy0` should be updated to `toolshed.g2.bx.psu.edu/repos/iuc/rgrnastar/rna_star/2.7.8a`

The workflow release number has been updated from 0.1 to 0.2.

### DIFF
--- a/workflows/transcriptomics/rnaseq-pe/CHANGELOG.md
+++ b/workflows/transcriptomics/rnaseq-pe/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## [0.2] 2022-11-02
 
 ### Automatic update
-- `toolshed.g2.bx.psu.edu/repos/iuc/rgrnastar/rna_star/2.7.8a+galaxy0` was updated to `toolshed.g2.bx.psu.edu/repos/iuc/rgrnastar/rna_star/2.7.8a`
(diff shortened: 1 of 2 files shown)


lldelisle:
What is strange is that in the file `.ga` it changed from galaxy0 to galaxy1...

lldelisle:
The tool toolshed.g2.bx.psu.edu/repos/iuc/rgrnastar/rna_star/2.7.8a+galaxy1 is not yet installed in usegalaxy.org.
I need to change the test case because it is too large and takes too long.

mvdbeek:
It's also installed now and passing.

lldelisle:
Yes I was wondering if I could change the input test data as I think the CI is a bit long...

lldelisle:
But maybe I can do it in a second time.

mvdbeek (approved):
That's up to you, there's no cost to having another version :)

lldelisle:
Reducing the size of input test did not change the running time...

PR #153, opened 2022-11-23
creation of workflow for plotting of contig/scaffold sizze and Nx for…

Delphine-L:
… one and 2 assemblies using gfastats outputs


mvdbeek:
Can you add the missing metadata ? https://github.com/galaxyproject/iwc/blob/main/workflows/README.md#add-required-metadata

wm75 (commented):
I think you can shorten the single output WF significantly by:

- making use of the latest Compute version
- replacing tail/cut/param value/add column with datamatsh/param value/compose expression

see my demo at https://usegalaxy.eu/u/wolfgang-maier/w/plot-gfastats-revised

This will also save a column in the computed table and you could save two more by making use of the new Compute tools Replace action instead of having it Append new columns.


Haven't looked at the two outputs WF, but maybe it can do similar things?

workflows/VGP-assembly-v2/VGP-Plot_gfastats_pair/Plot-gfastats-output---2-assemblies.ga
```diff
@@ -0,0 +1,1576 @@
+{
+    "a_galaxy_workflow": "true",
+    "annotation": "",
+    "format-version": "0.1",
+    "release": "0.1",
+    "name": "Plot gfastats output - 2 assemblies",
+    "steps": {
```
simleo, line 7: `license` and `creator` are missing

workflows/VGP-assembly-v2/VGP-Plot_gfastats/Plot-gfastats-output.ga
```diff
@@ -0,0 +1,604 @@
+{
+    "a_galaxy_workflow": "true",
+    "annotation": "",
+    "format-version": "0.1",
+    "release": "1.0",
+    "name": "Plot gfastats output",
```
simleo, line 6: `license` and `creator` are missing

PR #179, opened 2023-02-22
Add Pox Virus Illumina Amplicon workflow

kciy:
Proposing a workflow that can be used for analysis of pox virus genomes that are sequenced as half-genomes in a tiled-amplicon approach; here for Lumpy Skin Disease Virus.

Could restructure and nest this workflow with SARS-CoV-2 workflows in a new folder `virology/` ?


wm75:
@mvdbeek testing is expected to fail currently because the WF uses new versions of ivar which I don't think have been installed on .org yet.

When updating on .org, there is a catch to be aware of:
the latest release (1.3.2) of ivar contains a major bug in ivar trim, which makes this particular tool very inefficient (https://github.com/andersen-lab/ivar/issues/147). On .eu we did not install this version of this one tool because of that, and you may want to do the same on .org.
The WF uses 1.3.2 for ivar consensus, but 1.3.1 (but a new wrapper version) for ivar trim for this reason.

wm75:
@TKlingstrom that's the new mapping-based WF for LSDV we've been discussing

mvdbeek:
> testing is expected to fail currently because the WF uses new versions of ivar which I don't think have been installed on .org yet.

tools come from the shed now

mvdbeek:
> Could restructure and nest this workflow with SARS-CoV-2 workflows in a new folder `virology/` ?

that'd be great, can you start by placing this particular workflow under virology, then in another PR we can move the SARS-CoV-2 workflows

workflows/virology/pox-virus-amplicon/README.md
```diff
@@ -0,0 +1,5 @@
+# Pox Virus Illumina Amplicon Workflow from half-genomes
+
+This workflow generates a consensus sequence from Illumina-sequenced ARTIC PE data using BWA-MEM for mapping and iVar for consensus sequence generation.
+
+It requires that the pox virus genome has been sequenced in two halves in two separate sequencing runs.
```
wm75, line 5:
```suggestion
# Pox Virus Illumina Amplicon Workflow for half-genomes sequencing data

This workflow generates consensus sequences from Illumina PE-sequenced ARTIC data of pox virus samples.

It requires that all samples have been sequenced in two halves in two separate sequencing runs, and utilizes this property to resolve the inverted terminal repeat (ITR) sequences of pox virus genomes.

The workflow uses BWA-MEM for mapping the reads from each half-genome sequencing run to a correspondingly masked version of the reference genome, merges the resulting two read mappings, and uses iVar for primer trimming and consensus sequence generation.

Conceptually, this workflow builds on https://github.com/iwc-workflows/sars-cov-2-pe-illumina-artic-ivar-analysis and adds the logic for the split genome mapping and merging of the results.
```

workflows/virology/pox-virus-amplicon/CHANGELOG.md
```diff
@@ -0,0 +1,5 @@
+# Changelog
+
+## [0.1]
+
+- Initial version of Pox Virus Illumina Amplicon workflow (iVar based) for half-genomes sequencing data
```
wm75, line 5:
```suggestion
# Changelog

## [0.1]

Initial version of Pox Virus Illumina Amplicon workflow (iVar based) for half-genomes sequencing data
```

wm75 (commented):
@kciy, great work!
I've suggested to extend the README a bit.

wm75:
@jonas-fuchs fyi :)

mvdbeek:
Screenshot 2023-02-22 at 10 48 28

that looks a bit like a tool error ?

wm75:
Yeah, that came up in local testing, but not on .eu where the WF was developed :(

mvdbeek:
Maybe that can be replaced with https://usegalaxy.eu/root?tool_id=toolshed.g2.bx.psu.edu/repos/iuc/samtools_merge/samtools_merge/1.15.1+galaxy0 ?

wm75:
That was one of our ideas, too, yes. Let me chekc whether that one preserves RG-IDs as nicely as the picard tool.

mvdbeek:
The problem I think is that if you run the tool containerized we don't run through the conda activate scripts that setup JAVA_JAR_PATH. That said samtools also has the `cat` command, which is much faster than merge and just doesn't do anything to the individual records.

wm75:
samtools merge seems like the right choice here because we want the headers merged and the reads sorted. Only waiting for the tool run to make it out of the queue on .eu :)

mvdbeek:
run it locally ? it should be pretty fast, and if you use `planemo test/serve --biocontainers --shed_tool_conf /tmp/shed_tool_conf.xml --shed_tool_path /tmp/shed_dir` you can reuse the tools that are getting pulled in.
We should totally add a `--use_cached_jobs` flag too :)

wm75:
Thanks @kciy. I fixed a few things that got broken with your tool replacement.
Asking for unique @PG lines in the header avoids duplication of those elements (the wrapper is not very user-friendly here so I did it wrong in my first attempt, too).

wm75 (approved):

workflows/virology/pox-virus-amplicon/pox-virus-half-genome-test.yml
```diff
@@ -0,0 +1,46 @@
+- doc: Testing Pox Virus Illumina Amplicon Workflow with Lumpy Skin Disease Virus half genome sequenced data
+  job:
+    Reference FASTA:
+      class: File
+      location: "https://www.ebi.ac.uk/ena/browser/api/fasta/AF325528.1?download=true"
+      filetype: fasta
+    Primer Scheme (BED file with Pool identifiers):
```
mvdbeek, line 7: Can we put the instructions into the step annotations ?

I think this looks a bit nicer as `Primer Scheme`.
Same for `PE Pool1 read collection for samples` which could be `PE Pool1` and `PE Pool2 read collection for samples` which could be `PE Pool2`.

workflows/virology/pox-virus-amplicon/README.md
```diff
@@ -0,0 +1,9 @@
+# Pox Virus Illumina Amplicon Workflow for half-genomes sequencing data
+
+This workflow generates consensus sequences from Illumina PE-sequenced ARTIC data of pox virus samples.
+
+It requires that all samples have been sequenced in two halves in two separate sequencing runs, and utilizes this property to resolve the inverted terminal repeat (ITR) sequences of pox virus genomes.
+
+The workflow uses BWA-MEM for mapping the reads from each half-genome sequencing run to a correspondingly masked version of the reference genome, merges the resulting two read mappings, and uses iVar for primer trimming and consensus sequence generation.
```
mvdbeek, line 7: I'm curious why you've not opted for bwa-mem2 ? Totally fine, but I thought that was a faster drop-in replacement

workflows/virology/pox-virus-amplicon/pox-virus-half-genome-test.yml
mvdbeek: Imagine we'll generate automated command line wrappers, much nicer to use `--primer-scheme`, `--pe-pool1` etc. I think that also looks better in the user interface

workflows/virology/pox-virus-amplicon/README.md
wm75: That's a really good question. As so often, the honest answer is that I didn't reevaluate conclusions drawn much earlier, but your comment prompted me to start this now. So, some current thoughts and data:
- my previous (but never experimentally verified) idea was that for small viral genomes the speed gain to be expected for bwa-mem2 would be on the very lower end of what's being reported for the human genome case.
  That's because the SA is much smaller to begin with and, typically, there should be less Smith-Waterman alignments required per read (because fewer candidate seed regions should exist in a small genome) - so optimizations in both areas would matter less. otoh, at the time, I was worried that there might still be bugs waiting to be discovered in bwa-mem2.
- now, that the bwa-mem2 code base has been stable for quite some time, it's a good time to evaluate the performance instead of just speculating
- from a limited number of runs (~10) of both tools on .eu so far it seems that:
  * the `main_mem` function of bwa-mem takes ~ 25% longer than bwa-mem2's (using AVX2 mode)
  * total wall run time of bwa-mem is, on average, ~30 % longer, but that's quite variable and possibly confounded by the fact that the bwa-mem wrapper uses samtools 1.6 for sorting, while bwa-mem2 uses 1.13 (in fact, only 50% of total run time is spent in `main_mem` for each tool and final bam sizes are a tiny bit larger for bwa-mem2)
  => so, yes, the speed gain is really at the lower edge of what's reported
- at the same time, bwa-mem2 jobs still require a bit more memory than bwa-mem ones (something like 50% more on average).

All in all, I would say using bwa-mem here like in other virological iwc WFs is not something that's a  dramatically wrong decision. However, we could consider to expand the testing to more datsets and job runs, and, if the outcome is convincing, switch all relevant WFs to bwa-mem2 at some point.

wm75: ah, and to put all those relative values into context we are talking about absolute values like:
- 15 (bwa-mem2) vs 19 secs spent in `main_mem`
- 30 vs 40 secs total job wall time
- 1 vs 1.5 GB memory

mvdbeek: Awesome, thanks for checking!

mvdbeek (approved):
Very cool, thanks a lot @wm75 @kciy !

PR #180, opened 2023-02-24
Move SARS-CoV-2 workflows to virology folder

kciy:
As briefly mentioned in #179, we want to nest all virology related workflows in the `virology/` folder. Are there more workflows that belong there?


wm75:
Hmm, not sure what the issue is with bamleftalign during linting.

lldelisle:
We are aware of this problem but this requires an update of the main toolshed...

lldelisle:
(See https://github.com/galaxyproject/galaxy/pull/14911)

PR #187, opened 2023-03-17
Updating workflows/virology/pox-virus-amplicon from 0.1 to 0.2

planemo-autoupdate:
Hello! This is an automated update of the following workflow: **workflows/virology/pox-virus-amplicon**. I created this PR because I think one or more of the component tools are out of date, i.e. there is a newer version available on the ToolShed.

* `toolshed.g2.bx.psu.edu/repos/iuc/qualimap_bamqc/qualimap_bamqc/2.2.2d+galaxy3` should be updated to `toolshed.g2.bx.psu.edu/repos/iuc/qualimap_bamqc/qualimap_bamqc/2.2.2c+galaxy1`
* `toolshed.g2.bx.psu.edu/repos/iuc/ivar_trim/ivar_trim/1.3.1+galaxy6` should be updated to `toolshed.g2.bx.psu.edu/repos/iuc/ivar_trim/ivar_trim/1.3.2+galaxy0`

The workflow release number has been updated from 0.1 to 0.2.

### DIFF
--- a/workflows/virology/pox-virus-amplicon/CHANGELOG.md
+++ b/workflows/virology/pox-virus-amplicon/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## [0.2] 2023-03-17
+
+### Automatic update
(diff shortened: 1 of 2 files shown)


wm75 (commented):
None of these two tool "updates" is desirable.
The first would mean downgrading, the second would update to a version suffering from a drastic [performance regression](https://github.com/andersen-lab/ivar/issues/147).

lldelisle:
@wm75,
For info, the log says that the qualimap_bamqc has been downgraded but in the workflow it is not. This is a problem we already observed with @mvdbeek

PR #194, opened 2023-06-09
[WIP] scRNA-seq Duboule lab workflows

lldelisle:
Hi,
I just upload here the workflows we are using in the lab for scRNA-seq from 10X:
- For classical 10x (using the subworkflow reorganize STAR output I don't know if I really need to put it into the repo).
- For cellPlex 10x (multiplexing) which is using the workflow above as subworkflow.
For the moment I am super busy with other stuff so there is no README, no test etc.


lldelisle:
@astrovsky01

lldelisle:
Moved to #314 and #315

PR #195, opened 2023-06-09
Three workflows to get a 'confident' narrowPeak from 2 replicates

lldelisle:
Hi,
No time to make README and tests for the moment but I think it can be useful.
I would like to improve it because for the moment it only works with a collection of 2 and ideally I would love that it works with a nested collection so we can run it on different samples in parallel and more than 2 would be also super cool.


lldelisle:
Read for review.

mvdbeek (approved):
This is so cool! I think my only comment would be to rename peaks-consensus to consensus-peaks.

lldelisle:
Your english is better than mine. I will change.

PR #200, opened 2023-07-15
Added new workflow to polish assembly

AnnaSyme:
Added to folder "genome-assembly", in subfolder "polish-with-long-reads"


AnnaSyme:
ping @mvdbeek  this is ready to go if you wouldn't mind having a look?

mvdbeek:
Looks great, my only suggestion is to the minimap profile parameter to restrict the values users can select (or provide a list of options, if some of them don't make sense)
Screenshot 2023-08-16 at 10 14 37

AnnaSyme:
thanks @mvdbeek this is a good idea. I want to provide a default value but also a list of suggested values. However is this possible? When I do put in the default, but also the suggested values, I can't see them in the drop-down list when I go to run the workflow. If I have no default value, I can see all the suggested values. Is there a way around this?

mvdbeek:
Kind of, the default value only works for optional parameters, so what would work (but is a little clunky) is when you define the possible parameters, you can put the default first.
Screenshot 2023-08-23 at 10 41 21

AnnaSyme:
thanks have done this for these suggested values @mvdbeek

mvdbeek:
Do you need to move Galaxy-Workflow-Assembly_polishing_with_long_reads.ga to Assembly_polishing_with_long_reads.ga ? There are now two workflows in here, and only one has a test.

AnnaSyme:
Thanks @mvdbeek - I've removed the older workflow and re-named the newer one - does that look ok?

workflows/genome-assembly/polish-with-long-reads/.dockstore.yml
```diff
@@ -0,0 +1,11 @@
+version: 1.2
+workflows:
+- name: main
+  subclass: Galaxy
+  publish: true
+  primaryDescriptorPath: /Assembly-polishing-with-long-reads.ga
```
mvdbeek, line 6: You'd have to update this as well

AnnaSyme: Should I update the version?

mvdbeek: No, it seems your workflow is called `Assembly_polishing_with_long_reads.ga` now, you either need to rename the workflow to `Assembly-polishing-with-long-reads.ga` or change this to `Assembly_polishing_with_long_reads.ga`. I think my preference would be to rename the workflow back to `Assembly-polishing-with-long-reads.ga`

workflows/genome-assembly/polish-with-long-reads/Assembly-polishing-with-long-reads.ga
```diff
@@ -0,0 +1,532 @@
...
+    "a_galaxy_workflow": "true",
+    "annotation": "Racon polish with long reads, x4",
+    "creator": [
+        {
+            "class": "Person",
+            "identifier": "https://orcid.org/0000-0002-9906-0673",
+            "name": "Anna Syme"
+        }
+    ],
+    "format-version": "0.1",
```
mvdbeek, line 11:
```suggestion
    "format-version": "0.1",
    "release": "0.1",
```

mvdbeek (approved):
very cool, I'll merge when the tests are passing!

bgruening:
:tada: :tada: :tada: :tada: :tada: :tada:

PR #201, opened 2023-09-01
fix dockstore for HiC

lldelisle:


lldelisle:
I modified versions in each workflow... I don't know if it was needed

mvdbeek (approved):

PR #203, opened 2023-09-05
Updating workflows/epigenetics/atacseq from 0.5 to 0.6

planemo-autoupdate:
Hello! This is an automated update of the following workflow: **workflows/epigenetics/atacseq**. I created this PR because I think one or more of the component tools are out of date, i.e. there is a newer version available on the ToolShed.

* `toolshed.g2.bx.psu.edu/repos/devteam/picard/picard_MarkDuplicates/2.18.2.4` should be updated to `toolshed.g2.bx.psu.edu/repos/devteam/picard/picard_MarkDuplicatesWithMateCigar/2.18.2.3`

The workflow release number has been updated from 0.5 to 0.6.

### DIFF
--- a/workflows/epigenetics/atacseq/CHANGELOG.md
+++ b/workflows/epigenetics/atacseq/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## [0.6] 2023-09-05
+
+### Automatic update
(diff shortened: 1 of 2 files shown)


lldelisle:
Looks like awk and grep tools have changed hashtag but not version.

PR #204, opened 2023-09-05
Updating workflows/epigenetics/cutandrun from 0.4 to 0.5

planemo-autoupdate:
Hello! This is an automated update of the following workflow: **workflows/epigenetics/cutandrun**. I created this PR because I think one or more of the component tools are out of date, i.e. there is a newer version available on the ToolShed.

* `toolshed.g2.bx.psu.edu/repos/devteam/picard/picard_MarkDuplicates/2.18.2.4` should be updated to `toolshed.g2.bx.psu.edu/repos/devteam/picard/picard_MarkDuplicatesWithMateCigar/2.18.2.3`

The workflow release number has been updated from 0.4 to 0.5.

### DIFF
--- a/workflows/epigenetics/cutandrun/cutandrun.ga
+++ b/workflows/epigenetics/cutandrun/cutandrun.ga
@@ -10,7 +10,7 @@
     ],
     "format-version": "0.1",
     "license": "MIT",
-    "release": "0.4",
+    "release": "0.5",
     "name": "CUTandRUN",
     "steps": {
         "0": {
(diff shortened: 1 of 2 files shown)


lldelisle:
Seems to be a change in grep hash without version change.

workflows/epigenetics/cutandrun/cutandrun.ga
```diff
@@ -676,7 +671,7 @@
             },
             "tool_id": "toolshed.g2.bx.psu.edu/repos/bgruening/text_processing/tp_grep_tool/1.1.1",
             "tool_shed_repository": {
-                "changeset_revision": "ddf54b12c295",
+                "changeset_revision": "d698c222f354",
```
mvdbeek, line 674: This is indeed the latest version, and you'd get this version when installing the workflow, so it's probably not worth creating a new version if there's no actual change.

lldelisle: I agree. However, I think we need to keep this PR open so autoupdate does not create a new one next time.

mvdbeek: Hmm, the autoupdate bot needs some love for this sort of things. Other bots allow you to close a PR under some conditions and will not re-open a new PR if there are no further changes. Or maybe let's try and see if it really opens a new PR with no new changes ?

lldelisle: OK

PR #212, opened 2023-09-18
VGP workflow 3

mvdbeek:


mvdbeek:
hifiasm is still failing with exit code 137, which should be OOM ...

mvdbeek:
> For small genomes, use -f0 to disable the initial bloom filter which takes 16GB memory at the beginning. For genomes much larger than human, applying -f38 or even -f39 is preferred to save memory on k-mer counting.

PR #217, opened 2023-09-22
fix bug in ATAC-seq workflow

lldelisle:
I am sorry. I forgot a step in 0.5.0...


mvdbeek (approved):
No worries, that's what we have versions for :)

PR #225, opened 2023-10-03
Add baredSC

lldelisle:


lldelisle:
As it is related to single-cell, it may be a good idea to put all single-cell RNA-seq into a category different from transcriptomics.

mvdbeek:
Agreed, we may not expect them there.

lldelisle:
I thought https://github.com/galaxyproject/planemo/pull/1371 was fixing the following case but apparently I was wrong.
I have a collection of outputs datasets.
All datasets have the same name as they come from the same input (by the way my output naming was not good).
When planemo tests the assertion, it applies the test to the last output with the same name.

lldelisle:
@PierreOsteil FYI

lldelisle:
Reviews are welcome

workflows/scRNAseq/baredsc/README.md
```diff
@@ -0,0 +1,29 @@
+# BaredSC Workflows
+
+These workflows allow to run a baredSC analysis in a single click. It uses models from 1 to N Gaussians and combine them. It uses the logNorm scale, 100 bins for 1 dimension and 25 bins on each axis in 2 dimensions.
+
+## Inputs dataset
+
+- Both workflows need a tabular dataset where each row is a cell. The tabular needs to have a header line with column names. There must be at least two columns: 'nCount_RNA' and another one with the counts for the gene(s) of interest.
```
mvdbeek, line 7: Where would I find / how would I produce such a file ?

lldelisle: Yes I should add it into the README. I will do it tomorrow.

lldelisle:
@mvdbeek, tell me if the description is clear enough. Also I could add the same workflows with Scanpy inputs however the problem is that I do not know where to find scanpy inputs to make the tests...

mvdbeek (approved):
Awesome, this looks great!

PR #226, opened 2023-10-04
Bump pyyaml version

simleo:
With Python 3.10 or newer, [PyYAML 5.4 fails to install](https://github.com/yaml/pyyaml/issues/724) due to a problem with Cython. This is not affecting the CI now since the github workflows are using Python 3.7, but it will if the Python version is bumped in the future (also, it affects anyone who wants to install locally with a recent Python version).


mvdbeek:
Thank you @simleo!

PR #231, opened 2023-10-16
Add RepeatElement workflow

rlibouba:


abretaud:
Thanks @rlibouba! not sure we want to submit this one to IWC, we'll need to test it with biohackathon data before I think

mvdbeek:
I've converted it to draft for now, but let me when it's ready or if you need help

PR #251, opened 2023-11-03
Update README.md

nekrut:


mvdbeek:

https://github.com/galaxyproject/iwc/actions/runs/6771946663

mvdbeek:
Can we go through the normal review process here ? I know these are minimal changes for the paper, but we now have divergent text in the workflow and the readme. I would also rather not include manually generated screenshots in the markdown, since that prevents us from automatically dumping the readme content back into the workflow's description field.

PR #252, opened 2023-11-06
Make tests more lenient for VGP9 that is failing in deployment

Delphine-L:


workflows/VGP-assembly-v2/Assembly-decontamination-VGP9/Assembly-decontamination-VGP9-tests.yml
```diff
@@ -9,6 +9,6 @@
     full contaminant + mito scaffold list:
       file: test-data/full contaminant + mito scaffold list.txt
       compare: diff
-      lines_diff: 1
+      lines_diff: 2
```
mvdbeek, line 12:
```suggestion
      sort: true
```

PR #254, opened 2023-11-07
added tag filtering for wf4. rearranged input ordering for wf1 and wf4

nekrut:
Adds tag filtering to wf4 to automatically select necessary genomescope and meryl outputs of wf1


mvdbeek (approved):
Thank you!

PR #259, opened 2023-11-07
added tags to hifiasm outputs so they can be easily fed to purge_dups…

nekrut:
… workflow


mvdbeek:

https://github.com/galaxyproject/iwc/actions/runs/6791289751

PR #317, opened 2023-12-05
Fix tests

mvdbeek:
Some more cosmetic fixes (`that: "has_text"` -> `has_text`), but also fixes a lot of untested broken things here.
Mostly a WIP to show the impact of https://github.com/galaxyproject/galaxy/pull/17128.

The list of asserts needs to be added to galaxy-tool-util for tests to pass.


lldelisle:
This is a bit scary how many tests were malformed, no?

mvdbeek:
Yes, confirming that we really needed to do this. It's too easy to get it wrong without linting, schema and very minimal docs.

lldelisle:
Thanks for doing it.

PR #321, opened 2024-01-11
add a workflow to connect optional workflow inputs to non-optional tool inputs

bernt-matthias:
Using this workflow as sub-workflow should simplify workflows. I think that having this might be cool irrespective of if we do https://github.com/galaxyproject/tools-iuc/pull/5713 + https://github.com/galaxyproject/galaxy/pull/17271 -- which do more or less the same .. but still its two tools (which have nothing to do with the actual WF vs a single subworkflow witha clearly described purpose)

Screenshot from 2024-01-11 23-04-38


mvdbeek:
We should be able to infer the input data type, that's probably not that hard to do, as we do that with other outputs as well.

bernt-matthias:
You mean in the pick value tool? Question is how to set this? Can we do this in the ECMAS script? Is there an example / docs (I'm really missing docs on expression tools .. is there any)?

mvdbeek:
Yes, format_source should just do it in https://github.com/galaxyproject/tools-iuc/blob/main/tools/pick_value/pick_value.xml#L149

bernt-matthias:
Will try. But syntax and functionality-wise I'm not entirely sure how to address a dataset that is nested in two conditionals and a repeat (which might have some null datasets) :)

bernt-matthias:
Syntax-wise `style_cond|type_cond|pick_from_0|value` will work. But this would take the format always from the first repeat element. For the workflow (editor) this might already be sufficient to determine the formats (even if the dataset is left empty in the first repeat), but I'm wondering what happens then during tool execution if the first dataset is given / not. I have the feeling that it might actually  work .. but need a bit of time for testing.


Wondering what happens if the fir

PR #322, opened 2024-01-11
Comment on PR with result of workflow test

lldelisle:
This is highly inspired from [this post](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/).
Contrarily to current code it works for PR from forks.

I deployed it on my fork and the issue I see is that when the PR workflow does not generate a combined test then the action Comment on the pull request fails because there is no artifact. See https://github.com/lldelisle/iwc/actions/runs/7492377134/job/20395726430

But I don't know how to introduce a if within the script to prevent filtering artifact if it is empty.

### DIFF
--- a/.github/workflows/comment_on_pr.yml
+++ b/.github/workflows/comment_on_pr.yml
@@ -9,15 +9,15 @@ on:
       - completed
 
 jobs:
-  upload:
+  download_and_comment:
     runs-on: ubuntu-latest
     if: github.event.workflow_run.event == 'pull_request'
     steps:
       - name: 'Download artifact'
-        uses: actions/github-script@v3.1.0
+        uses: actions/github-script@v7
         with:
           script: |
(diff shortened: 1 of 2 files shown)


bernt-matthias:
> when the PR workflow does not generate a combined test then the action Comment on the pull request fails because there is no artifact

In which cases there is no combined test?

lldelisle:
For example, this PR do not generate a combined test... All PRs that do not modify workflows...

mvdbeek:
that sounds ok to me ? and we'd also not want to get a comment if the PR passed ... i am in fact worried about getting spammed, considering that I already get a notification for every failed run of my own PRs

mvdbeek (approved):

lldelisle:
May I merge?

mvdbeek:

https://github.com/galaxyproject/iwc/actions/runs/7503040653

lldelisle:
:+1:

PR #358, opened 2024-03-11
VGP9 - fix : regex was wrong and causing headers to be modified

Delphine-L:


workflows/VGP-assembly-v2/Assembly-decontamination-VGP9/Assembly-decontamination-VGP9-tests.yml
```diff
@@ -11,4 +11,4 @@
       compare: diff
       sort: true
     final decontaminated assembly:
-     checksum: 'sha1$8a413a18525c11fc0da27c07992b968490e0c1b9'
\ No newline at end of file
+     checksum: 'sha1$70a3351156b45e1961e7b986f330527a175f9794'
```
mvdbeek, line 14: Can you include a test that would have failed previously ? The checksum doesn't tell us anything about the contents of the file.

mvdbeek (approved):

PR #395, opened 2024-04-16
Simplify parallel accession download workflow

bernt-matthias:
by passing ID file directly

we could also add a parameter allowing to change the number of IDs per chunk

### DIFF
--- a/workflows/data-fetching/parallel-accession-download/CHANGELOG.md
+++ b/workflows/data-fetching/parallel-accession-download/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## [0.1.12] 2024-04-16
+
+### Skip param file generation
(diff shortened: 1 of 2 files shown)


workflows/data-fetching/parallel-accession-download/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## [0.1.12] 2024-04-16
+
+### Skip param file generation
+
+- instead pass ID file directly
```
mvdbeek, line 7:
```suggestion
### Changed

- Remove unnecessary param_value_from_file step and directly pass file with identifiers to fasterqdump
```

mvdbeek (approved):
Nice, thank you!

PR #411, opened 2024-05-01
Version 1.0 of consensus-peaks

lldelisle:


lldelisle:
Ready for review

workflows/epigenetics/consensus-peaks/CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## [1.0] 2024-05-01
+
+The workflow changed as it accepts now more than 2 replicates and has a new parameter to define what is the minimum number of replicates where the summit should be found.
+The workflow for ATAC and CUT&RUN changed the input datatype as now a BAM is required instead of a BED.
```
mvdbeek, line 6:
```suggestion
### Changed
The workflow changed as it accepts now more than 2 replicates and has a new parameter to define what is the minimum number of replicates where the summit should be found.
The workflow for ATAC and CUT&RUN changed the input datatype as now a BAM is required instead of a BED.
```

mvdbeek (approved):
That's a cool improvement, thank you @lldelisle!

lldelisle:
Thanks. I think it is our first workflow with comments...

PR #438, opened 2024-06-05
Add 'Bacterial genome assembly' workflow

clsiguret:
This workflow comes from the ABRomics project and can be used for any bacterial genomic analysis:
**Bacterial Genome assembly for paired end data**

We followed the recommendation to publish the workflow (licence, author list and email, steps annotations...)

- [x] Workflow imported from [usegalaxy.fr](https://usegalaxy.fr/) using `planemo workflow_test_init` to use a valid invocation
- [x] Workflow tests passed using `planemo test` command on the [usegalaxy.fr](https://usegalaxy.fr/) and [usegalaxy.eu](https://usegalaxy.eu/) servers
- [x] Readme.md written
- [x] CHANGELOG.md written
- [x] .dockstore.yml written

PR #439, opened 2024-06-05
Add 'AMR gene detection' workflow

clsiguret:
This workflow comes from the ABRomics project and can be used for any bacterial genomic analysis:
AMR gene detection in an assembled bacterial genome

This workflow uses 2 databases which must be available on http://datacache.galaxyproject.org/:
- [amrfinderplus_database](http://datacache.galaxyproject.org/indexes/amrfinderplus-db/) (**AMRFinderPlus**): amrfinderplus_V3.12_2024-05-02.2
- abricate_database (**ABRicate**): vfdb

We followed the recommendation to publish the workflow (licence, author list and email, steps annotations...)
- [x] Workflow imported from [usegalaxy.fr](https://usegalaxy.fr/) using `planemo workflow_test_init` to use a valid invocation
- [x] Workflow tests passed using `planemo test` command on the [usegalaxy.fr](https://usegalaxy.fr/) server
- [x] Readme.md written
- [x] CHANGELOG.md written
- [x] .dockstore.yml written


workflows/bacterial_genomics/amr_gene_detection/CHANGELOG.md
```diff
@@ -0,0 +1,5 @@
+# Changelog
+
+## [0.1] - 05-06-2024
```
wm75, line 3: doesn't match release (1.0) stated in WF file

workflows/bacterial_genomics/amr_gene_detection/README.md
```diff
@@ -0,0 +1,30 @@
+# AMR gene detection in an assembled bacterial genome subworkflow (v1.0)
+
+This subworkflow use assembled bacterial genome fasta files and executes the following steps:
```
wm75, line 3: Not sure I would call it a subworkflow since it doesn't make any assumptions about how the bacterial consensus sequence was generated -> would work on any fasta genome you provide, wouldn't it?

workflows/bacterial_genomics/amr_gene_detection/amr_gene_detection.ga
```diff
@@ -0,0 +1,683 @@
+{
+    "a_galaxy_workflow": "true",
+    "annotation": "Analysis of assembly genome to provide antimicrobial gene detection",
```
wm75, line 3: how about: "Antimicrobial resistance gene detection from assembled bacterial genomes"?

wm75:
Comments by @mvdbeek and me about step annotations and workflow tags in #438 and #437 also apply here.

wm75:
I understand that you're planning to use this as a subworkflow, but have you considered expanding it to work on a **collection** of fasta consensus sequences, i.e., on the outputs of several runs of upstream WFs?
Seems to be a good candidate for such parallel processing of multiple samples (and you could even extend it later to produce some aggregate reports).

workflows/bacterial_genomics/amr_gene_detection/CHANGELOG.md
clsiguret: Thanks !

workflows/bacterial_genomics/amr_gene_detection/README.md
clsiguret: I agree with you, so I changed it to "workflow" and clarified that it is possible to use any fasta file :smile:

workflows/bacterial_genomics/amr_gene_detection/amr_gene_detection.ga
clsiguret: Thanks for the suggestion !

clsiguret:
> I understand that you're planning to use this as a subworkflow, but have you considered expanding it to work on a **collection** of fasta consensus sequences, i.e., on the outputs of several runs of upstream WFs? Seems to be a good candidate for such parallel processing of multiple samples (and you could even extend it later to produce some aggregate reports).

I understand your remark. Indeed it can be a workflow in its own right, and for the moment I prefer not to use a collection (even if it would be better). Maybe for a new version of the workflow 😄

wm75:
> > I understand that you're planning to use this as a subworkflow, but have you considered expanding it to work on a **collection** of fasta consensus sequences, i.e., on the outputs of several runs of upstream WFs? Seems to be a good candidate for such parallel processing of multiple samples (and you could even extend it later to produce some aggregate reports).
> 
> I understand your remark. Indeed it can be a workflow in its own right, and for the moment I prefer not to use a collection (even if it would be better). Maybe for a new version of the workflow 😄

That's fine with me then.

bebatut (approved):

PR #445, opened 2024-06-19
Add BREW3R

lldelisle:
BREW3R stands for **B**ulk **R**NA-seq **E**vidence-based **W**orkflow for **3**' UTR **R**eannotation. 

This workflow enables extending an existing gtf downloaded on a public website, like Ensembl, Genecode or UCSC, using *de novo* gene annotation with StringTie on full length bulk RNA-seq.


workflows/transcriptomics/brew3r/BREW3R-tests.yml
```diff
@@ -0,0 +1,29 @@
...
+        path: test-data/GSM4196550_subset.bam
+      - class: File
+        identifier: GSM4196551
+        path: test-data/GSM4196551_subset.bam
+      - class: File
+        identifier: GSM4196552
+        path: test-data/GSM4196552_subset.bam
+    strandedness: 'stranded - reverse'
+    minimum coverage: '10'
+    minimum FPKM for merge: '1.0'
```
mvdbeek, line 22:
```suggestion
    minimum coverage: 10
    minimum FPKM for merge: 1.0
```
did you get this through test generation or did you add this manually ? It doesn't matter for tests now, but I hope that in the future we can check the types :)

mvdbeek (dismissed):
Wow, that's such a cool workflow 😍 ! Thank you!

workflows/transcriptomics/brew3r/BREW3R-tests.yml
lldelisle: I did `planemo workflow_test_init --from_invocation` and the parameters format was super strange:
```
    strandedness: stranded - reverse
    minimum coverage: '10'
    minimum FPKM for merge: '1.0'
```
I thought it would be better to put quotes around the strandedness and I wonder what to do with the float/int values.

mvdbeek: yaml infers types as best as it can, so the quotes are not necessary if the only type it can be is a string.
quotes around native yaml types however turn those values into strings.

```yaml
a: 1  # int
b: "1"  # string
c: a string  # string
d: "a string"  # string, same as c
e: "this: must have quotes"  # without quotes this would be a syntax error. 
```

mvdbeek (approved):

workflows/transcriptomics/brew3r/BREW3R-tests.yml
lldelisle: Thanks

PR #448, opened 2024-06-23
First attempt for more linting

lldelisle:
I think this is needed. I am not sure this is the good way. I don't even know if it works. I would need to test on my fork before we merge this.

### DIFF
--- a/.github/workflows/workflow_test.yml
+++ b/.github/workflows/workflow_test.yml
@@ -44,6 +44,14 @@ jobs:
         mode: lint
         workflows: true
         repository-list: "${{ needs.setup.outputs.repository-list }}"
+    - name: IWC specific lint
+      run: |
+        for REPO in ${{ needs.setup.outputs.repository-list }}; do
+          echo $REPO
+          cd $REPO
+          python3 scripts/lint_repo.py
+          cd -
+        done
 
   test:
     name: Test workflows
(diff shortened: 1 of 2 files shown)


.github/workflows/workflow_test.yml
```diff
@@ -44,6 +44,14 @@ jobs:
         mode: lint
         workflows: true
         repository-list: "${{ needs.setup.outputs.repository-list }}"
+    - name: IWC specific lint
+      run: |
+        for REPO in ${{ needs.setup.outputs.repository-list }}; do
```
lldelisle, line 49: This does not work. When the repository-list is more than one it is replaced by the different values separated by back to line character, for example:
```
for REPO in workflows/blabla/blibli
workflow/blabla/blublu; do
```
And this is not a valid bash command.

.github/workflows/workflow_test.yml
```diff
@@ -44,6 +44,14 @@ jobs:
...
+          echo $REPO
+          cd $REPO
+          python3 scripts/lint_repo.py
```
lldelisle, line 52: This is not working because the path `scripts/lint_repo.py` is not working anymore once you have done `cd` but I do not know where can I find the variable that contains the path where the github repo has been checked out.

scripts/lint_repo.py
```diff
@@ -0,0 +1,97 @@
+# This script will check that the directory
```
mvdbeek, line 1: This should go directory to planemo, I think

lldelisle: You mean directly?

mvdbeek: Yeah, `workflow_lint` should do this.

PR #455, opened 2024-06-28
Update instructions to find the invocation id

lldelisle:

PR #447, opened 2024-06-23
Improve README on workflows

lldelisle:
Hi there,
I added a part about multi-workflow directories (@bernt-matthias if you can have a look, would be great).
I added a part about the autoupdate and I think we need to change the strategy but I think it is good that it is described here.

PR #461, opened 2024-06-28
Fix README

lldelisle:
Fix problems introduced with #455 #447.


mvdbeek:

https://github.com/galaxyproject/iwc/actions/runs/9711170008

PR #510, opened 2024-08-19
Updating workflows/VGP-assembly-v2/Purge-duplicate-contigs-VGP6 from 0.4 to 0.5 

gxydevbot:
Hello! This is an automated update of the following workflow: **workflows/VGP-assembly-v2/Purge-duplicate-contigs-VGP6**. I created this PR because I think one or more of the component tools are out of date, i.e. there is a newer version available on the ToolShed.

* `toolshed.g2.bx.psu.edu/repos/bgruening/text_processing/tp_cut_tool/9.3+galaxy1` should be updated to `toolshed.g2.bx.psu.edu/repos/bgruening/text_processing/tp_cut_tool/9.3+galaxy2`

The workflow release number has been updated from 0.4 to 0.5.


lldelisle:
Hi @Delphine-L ,
The .dockstore.yml file has indeed no 'testParameterFiles' set:
https://github.com/galaxyproject/iwc/blob/39029a8b856dc1ca36b914543e724bc0eb9b2d40/workflows/VGP-assembly-v2/Purge-duplicate-contigs-VGP6/.dockstore.yml
So while the test.yml exists it is not used and not tested in the CI.
This linting part was not in the CI previously that's why it was not detected before.
Locally you need to run:
```
planemo workflow_lint --iwc workflows/VGP-assembly-v2/Purge-duplicate-contigs-VGP6
```
To be able to reproduce the error.
Cheers,

Delphine-L (approved):

PR #520, opened 2024-09-18
Quality and contamination control workflow update from 1.1.1 to 1.1.2

clsiguret:
Close #495 
Add Recentrifuge database in input parameter + change tool versions for ToolDistillator

Thanks


wm75:
Nice!

Can you add a note also to the changelog that you updated the tool versions manually in the tooldistillator params?
Increases the chances that we will remember to do so on the next update.

clsiguret:
Thanks @wm75 :smile:

wm75 (approved):
Can't think of a better way to address the version issue atm, but good to be aware of it at least.

PR #531, opened 2024-09-24
Fix test syntax issues passing

lldelisle:
See #530

### DIFF
--- a/workflows/VGP-assembly-v2/kmer-profiling-hifi-VGP1/kmer-profiling-hifi-VGP1-tests.yml
+++ b/workflows/VGP-assembly-v2/kmer-profiling-hifi-VGP1/kmer-profiling-hifi-VGP1-tests.yml
@@ -12,22 +12,18 @@
     Ploidy: 1
   outputs:
     GenomeScope linear plot:
-      class: File
       file: test-data/GenomeScope_Linear_plot.png
       compare: sim_size
       delta: 10000
     GenomeScope log plot:
-      class: File
       file: test-data/GenomeScope_Log_plot.png
       compare: sim_size
       delta: 10000
     GenomeScope transformed linear plot:
-      class: File
       file: test-data/GenomeScope_Transformed_linear_plot.png
       compare: sim_size
       delta: 10000
     GenomeScope transformed log plot:
-      class: File
       file: test-data/GenomeScope_Transformed_log_plot.png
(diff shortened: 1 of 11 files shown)


lldelisle:
Same question as #533 , I don't know if these changes require a bump version.

mvdbeek:
No, if we're only changing the tests I don't think we need to do that. It doesn't affect the workflow itself.

workflows/VGP-assembly-v2/kmer-profiling-hifi-VGP1/kmer-profiling-hifi-VGP1-tests.yml
```diff
@@ -12,22 +12,18 @@
     Ploidy: 1
   outputs:
     GenomeScope linear plot:
-      class: File
```
mvdbeek, line 15: @jmchilton do we like this ? It's a great union discriminator that would give us cleaner validation reports, I think ?

mvdbeek (approved):

workflows/VGP-assembly-v2/kmer-profiling-hifi-VGP1/kmer-profiling-hifi-VGP1-tests.yml
jmchilton: The inputs you could argue are files. These are datasets - definitely not files. But the way CWL uses the word ``File`` is also just a real problem for my literal brain. We've already given into the syntax so why not just continue I guess. It is hard to appreciate consistency and technical quality of something that is wrong 😆. We can make this required in the best practice schema I think 😿. Will you give me the ``that`` syntax on the same argument about having a good clean discriminator 😆?

mvdbeek: Sure, that's fine. I guess we're still at the stage of defining what works, and then next step consolidating into something we like ?

jmchilton: https://github.com/galaxyproject/galaxy/pull/18889

mvdbeek:

https://github.com/galaxyproject/iwc/actions/runs/11107437308

PR #533, opened 2024-09-24
Use correct syntax for tests of CUT&RUN tools

lldelisle:
@mvdbeek do this requires a version bump or the deployement will fail but at least it will be integrated to the next PR.


lldelisle:
Could someone review and merge?

mvdbeek:

https://github.com/galaxyproject/iwc/actions/runs/11109496850

PR #558, opened 2024-10-10
Updating workflows/genome-assembly/quality-and-contamination-control from 1.1.3 to 1.1.4

clsiguret:

Thanks :smiley:


lldelisle (approved):
This is a super good idea.

PR #560, opened 2024-10-11
 Updating workflows/bacterial_genomics/bacterial_genome_annotation from 1.1.5 to 1.1.6

clsiguret:

Thanks :smiley:


bebatut (approved):
Thanks!

clsiguret:
Thanks @bebatut :smiley:

PR #561, opened 2024-10-11
Updating workflows/bacterial_genomics/amr_gene_detection from 1.1.3 to 1.1.4

clsiguret:

Thanks :smiley:


bebatut (approved):
Thanks!

clsiguret:
Thanks  @bebatut :smiley:

bgruening:
Yeah, very cool!

PR #562, opened 2024-10-11
IWC showcase

dannon:
Very much still a work in progress but we want to go ahead and get this publishing for integrations w/ other servers and to collect any feedback and start polishing.

image

https://galaxyproject.github.io/iwc/


mvdbeek:

https://github.com/galaxyproject/iwc/actions/runs/11293246088

lldelisle:
This is already super cool.
Do you think you could do something like GTN is doing:
https://training.galaxyproject.org/training-material/topics/single-cell/tutorials/scatac-preprocessing-tenx/tutorial.html#admin-missing-tools
Where the tool_panel_section_label is based on usegalaxy.eu/org/au

PR #574, opened 2024-10-17
Testing if RNA-seq WF can run on different test data

pavanvidem:


mvdbeek:
https://github.com/galaxyproject/iwc/actions/runs/11387974876/artifacts/2069752040 has your results, I suppose you were expecting the test to fail ? The workflow did run to completion.

pavanvidem:
Yes, I was just checking whether the workflow can run to completion with this test data. I am aware that the tests would fail because I did not update the tests.

pavanvidem:
@mvdbeek @lldelisle the problem seems to be with the input collection. The last run has minimal changes to the existing tests. The workflow remains unchanged.  I added multiple paired fastqs to the collection and it failed. Do you see anything wrong with the tests?

lldelisle:
I relaunched the CI. Have you run it against org? Does it manage to create the collection? Can you share the history?

pavanvidem:
I am not sure whether it managed to create the collection. How to check this?

Here is the history: https://usegalaxy.org/u/videmp/h/rnaseq-pe-with-multi-fastq-collection

mvdbeek:
I said this at the last BRC meeting, if the test is cancelled like this you're running out of resources. So I would suggest downsampling your data and using a small reference genome.

pavanvidem:
Are those resources assigned per workflow or per individual tool run? 

The tests are passing if there is only one pair of FASTQs in the collection but are failing if there are 4. Please check my last commit. I just added 3 more FASTQs to the collection in the original tests. Nothing else was changed.

mvdbeek:
The VM has 8 gigs of ram, but one of your star jobs consumes 9.3 gigs of ram.

pavanvidem:
@mvdbeek Thanks for the numbers! Using chr4 annotations only should consume a bit less memory (maybe just under 8G). Let's see. 
But most of the memory footprint is from loading STAR index. Finding a new decent dataset takes a while. Hence, it would be nice if can havedm6 chr4 only STAR index. That should make tests faster and will be also useful for the tutorial. Is it feasible now so that we can have the workflow ready?

mvdbeek:
> Is it feasible now so that we can have the workflow ready?

No, sorry.

PR #581, opened 2024-10-24
Add new ChIP-Seq WF that handles replicates and controls

wm75:
Not quite ready yet, but time to bring online.

@lldelisle another variation of the theme :-)
Essentially, an automated version of the steps in https://training.galaxyproject.org/training-material/topics/epigenetics/tutorials/formation_of_super-structures_on_xi/tutorial.html#formation-of-the-super-structures-on-the-inactive-x


lldelisle:
Hi,
Looks great indeed and too complex to be merged with the existing one.

wm75:
> Hi, Looks great indeed and too complex to be merged with the existing one.

That's what I thought, too, even though I borrowed a lot from your WFs (like the entire backbone of the analysis, and the averaging across replicates), so if you want to, I can add you as a WF author.

You may want to have a look at using fastp instead of cutadapt, which makes it possible to have optional adapter sequences and the updated samtools view step I'm using. Those could be integrated easily into your WF I guess.

lldelisle:
Yes, I also think it would be beneficial to use fastp instead of cutadapt in all my workflows.

pavanvidem:
@wm75 it seems there is no 'Calculate numeric parameter value' tool on .org.

wm75:
Funny, it's an iuc tool: https://github.com/galaxyproject/tools-iuc/tree/main/tools/calculate_numeric_param
Probably just oversight @mvdbeek?

bgruening:
Please add it here: https://github.com/galaxyproject/usegalaxy-tools

workflows/epigenetics/chipseq-pe-with-replicates-controls/chipseq-pe-with-replicates-controls.ga
```diff
@@ -0,0 +1,3692 @@
...
+            },
+            "tool_state": "{\"advancedOpt\": {\"showAdvancedOpt\": \"yes\", \"__current_case__\": 1, \"binSize\": \"50\", \"sortRegions\": \"keep\", \"sortUsing\": \"mean\", \"averageTypeBins\": \"mean\", \"missingDataAsZero\": true, \"skipZeros\": true, \"minThreshold\": null, \"maxThreshold\": null, \"scale\": null, \"metagene\": false, \"transcriptID\": \"transcript\", \"exonID\": \"exon\", \"transcript_id_designator\": \"transcript_id\", \"blackListFileName\": {\"__class__\": \"RuntimeValue\"}}, \"custom_sample_labels_conditional\": {\"custom_labels_select\": \"No\", \"__current_case__\": 0}, \"mode\": {\"mode_select\": \"reference-point\", \"__current_case__\": 1, \"referencePoint\": \"center\", \"nanAfterEnd\": false, \"beforeRegionStartLength\": \"3000\", \"afterRegionStartLength\": \"3000\"}, \"multibigwig_conditional\": {\"orderMatters\": \"No\", \"__current_case__\": 0, \"bigwigfiles\": {\"__class__\": \"ConnectedValue\"}}, \"output\": {\"showOutputSettings\": \"no\", \"__current_case__\": 0}, \"regionsFiles\": [{\"__index__\": 0, \"regionsFile\": {\"__class__\": \"ConnectedValue\"}}], \"__page__\": null, \"__rerun_remap_job_id__\": null}",
```
wm75, line 3357: So the beforeRegionStartLength and the afterRegionStartLength here should not be hard-coded.
Initially I thought I could come up with a way to estimate them from the MACS2 peak region widths, but maybe we should just turn them into a WF param for now.

mvdbeek: Should we start with the hardcoded region length for tomorrow ?

workflows/epigenetics/chipseq-pe-with-replicates-controls/test-data/test_sample_sheet.tsv
```diff
@@ -0,0 +1,4 @@
+SRR5204807	Spt5	rep1	SRR5204809
```
mvdbeek, line 1: Can we not create this file dynamically by having the users provide sample and ip collections ? I don't hate this for running this on the command line, but a lot can and will go wrong here. You mentioned issues if control and ip numbers differ, but I think that would also still be the case with the sample sheet.

wm75: One of the issues is that with the sample sheet you can define the same sample to be the appropriate control for several samples in a replicate, or each sample can have its own control, or whatever. In the sample sheet this connection is made via columns 1 and 4, but you cannot do this flexibly with just collections (unless you rely on naming schemes of elements, which is even worse than the sample sheet approach).
Averaging across replicates is also harder and more error-prone without a sample sheet, which establishes nicely which samples represent the same condition across replicates (maybe look at the example from the README).

workflows/epigenetics/chipseq-pe-with-replicates-controls/chipseq-pe-with-replicates-controls.ga
wm75: yes, please :+1:

workflows/epigenetics/chipseq-pe-with-replicates-controls/test-data/test_sample_sheet.tsv
mvdbeek: Can you make it two input collections, and if you need to repeat a control .... you repeat the control ? I don't see the difficulty currently. I would currently not let users build nested collections.

mvdbeek: Replicates are determined by the order of the collection elements, not the identifier

wm75: Note that in the above sample A and B, for example, are not IPed sample and control, but two different IPs done on the same input.

mvdbeek: Can we say one contrast per analysis for a start ? If you want to use a sample sheet you can have a parent workflow align your inputs according to the sheet ? I think that would nicely encapsulate the complexity while maintaining a easy to use workflow for the common case of doing a single condition ?

wm75: Hmm, that could be a way forward, but not for tonight? Quite some rewiring, tests rewriting, and manual testing, plus readme rewrite that's required for this.
As I said before, both @lldelisle 's chipseq_pe and/or cutandrun WFs are a good starting point for tomorrow, too.

PR #582, opened 2024-10-25
first release of RNAseq DE analysis, filtering and plotting workflow

pavanvidem:
This workflow is downstream analysis of the existing rnaseq-pe and rnaseq-sr workflows. It takes 2 collections and performs differential expressin analysis, annotates the results table, filters and generates a valcano plot, a z-scores heatmap and a heatmap of normalized counts.


workflows/transcriptomics/rnaseq-de/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+## Inputs dataset
+
+- The workflow takes two collections of count files, each correcsponds to an experimental condition
+- Gene Annotaton: The same GTF file used for mapping and quantification. It is used to annotate the DESeq2 results table. Ideally, the GTF file should contain `gene_id`, `gene_biotype` and `gene_name` attributes.
+
+## Inputs values
+
+- Adjusted p-value threshold: All the genes with an adjusted p-value less than this value are considered as differentially expressed. With a value of 0.05, expect 5% of false positives in the differentially expressed genes list.
+- log2 fold change threshold: All the genes with an absolute fold cahnge (regarless of up or down regulation) more than this value are selected.
```
lldelisle, line 11:
```suggestion
- log2 fold change threshold: All the genes with an absolute fold change (regarless of up or down regulation) more than this value are selected.
```

lldelisle:
:+1: This looks good.

lldelisle:
Maybe you should write in the directory or in the workflow is only for 2 conditions.

pavanvidem:
> Maybe you should write in the directory or in the workflow is only for 2 conditions.

sure!

lldelisle:
One of my colleague proposed that the input could be a single collection and a replicate assignment like 1,1,1,2,2,2 and a batch assignment like 1,2,3,1,2,3. I don't know if this makes sense or a samples plan like @wm75 proposed in #581 (with less columns).
Maybe we could just add 3 general workflows: 
- split collection in 2 by samplesplan: one column is identifier, other column is group (new collection name)
- split collection in 2 by pattern: one collection contains identifiers which contains this pattern, another collection with identifiers which do not contain this pattern.
- split collection in 2 by comma-separated values: the user input would be `1,1,1,2,2,2` but this one is more error prone to me.

pavanvidem:
replicate assignments are fine because we know there are exactly 2 conditions. But I don't think it is possible to include batches unless we assume there are fixed number of batches. Where should we use these batches information? in multi-factor analysis?

wm75:
just simple two-conditions comparison might be good enough for this version.

wm75:
image

can you arrange the inputs in a more meaningful way @pavanvidem ?
(order is dependent on euclidian distance from the top-left corner :upside_down_face: )

wm75:
Also 0.0 doesn't seem like a good default for neither of  the thresholds.

lldelisle:
I think this version is fine. I just say that to make this more usable we should think about developing workflows to split the count collection into 2 (in another PR).

pavanvidem:
> Also 0.0 doesn't seem like a good default for neither of the thresholds.

These are non-optional  parameters of tools. I can only set a default if I make then optional and I have to use the dirty trick of connecting the steps and then making them optional with defaults :)

wm75:
> > Also 0.0 doesn't seem like a good default for neither of the thresholds.
> 
> These are non-optional parameters of tools. I can only set a default if I make then optional and I have to use the dirty trick of connecting the steps and then making them optional with defaults :)

The "better" trick is to use pick value after the optional param and if it's not set, pick the default value again :-)

lldelisle:
For me, 'Main factor counts' and 'Base factor counts' are not really meaningful.
For me a factor is for example 'Treatment' and then you have 2 levels: 'treated' and 'untreated'. One of them is the reference and I am not sure there is a consensus way to call the other level.
I propose: 'Counts in changed condition', 'Counts in reference condition'.

lldelisle:
> > > Also 0.0 doesn't seem like a good default for neither of the thresholds.
> > 
> > 
> > These are non-optional parameters of tools. I can only set a default if I make then optional and I have to use the dirty trick of connecting the steps and then making them optional with defaults :)
> 
> The "better" trick is to use pick value after the optional param and if it's not set, pick the default value again :-)

Why don't we simply put it optional with a default value?

pavanvidem:
> For me, 'Main factor counts' and 'Base factor counts' are not really meaningful. For me a factor is for example 'Treatment' and then you have 2 levels: 'treated' and 'untreated'. One of them is the reference and I am not sure there is a consensus way to call the other level. I propose: 'Counts in changed condition', 'Counts in reference condition'.

sounds better! I will change

wm75:
> For me, 'Main factor counts' and 'Base factor counts' are not really meaningful. For me a factor is for example 'Treatment' and then you have 2 levels: 'treated' and 'untreated'. One of them is the reference and I am not sure there is a consensus way to call the other level. I propose: 'Counts in changed condition', 'Counts in reference condition'.

Yes, the factor and factor levels should have better names. Alternatively, you could also use  input params there.

wm75:
> > > > Also 0.0 doesn't seem like a good default for neither of the thresholds.
> > > 
> > > 
> > > These are non-optional parameters of tools. I can only set a default if I make then optional and I have to use the dirty trick of connecting the steps and then making them optional with defaults :)
> > 
> > 
> > The "better" trick is to use pick value after the optional param and if it's not set, pick the default value again :-)
> 
> Why don't we simply put it optional with a default value?

https://matrix.to/#/%23galaxyproject_iwc%3Agitter.im/%24-cW9o37aiBvL760m03fDaXVkbprQjQ8-Qj9Qs7KSlb0?via=matrix.org&via=gitter.im :-)

pavanvidem:
- The count files are not so but no problem, I can move them to Zenodo. 
- I will make them invisible stuff except that needed

I am also building another 2 workflows (a paired and a single) in the same directory which will combine the quantification workflow and this one so that we will have complete DESeq2 workflows. But this PR can be merged before I add them.

lldelisle:
Would you mind to remove the test-data?

pavanvidem:
> Would you mind to remove the test-data?

sure, forgot :)

lldelisle (approved):
Looks good to me. Thank you so much.
@mvdbeek you want to review?

Issue #583, opened 2024-10-25
Workflows to write to split collection into 2


lldelisle:
We could add 3 general workflows:
- split collection in 2 by samplesplan: one column is identifier, other column is group (new collection name)
- split collection in 2 by pattern: one collection contains identifiers which contains this pattern, another collection with identifiers which do not contain this pattern.
- split collection in 2 by comma-separated values: the user input would be 1,1,1,2,2,2 but this one is more error prone to me.

See #582

wm75:
Yes, but what would be even better would be to have Apply Rules populate additional columns from an optional tabular input.
This way you could freely re-arrange collections based on metadata conatined within a dedicated file.

lldelisle:
Yes but Apply Rule is not so easy for most of users. If @pavanvidem or @wm75 are not working on this. I will do it.

pavanvidem:
I am not currently working on this. 

If we have some usecases of such workflows, we should implement it. If it is only for DESeq2 tool, then I think we should invest time in updating the tool.

lldelisle:
I regularly need to split collection based on pattern so I implemented it in #593

PR #593, opened 2024-11-08
Split collections into 2

lldelisle:
See #583

### DIFF
--- a/workflows/data-manipulation/split-collection/Split_collection_by_pattern_in_identifiers.ga
+++ b/workflows/data-manipulation/split-collection/Split_collection_by_pattern_in_identifiers.ga
@@ -1,0 +1,243 @@
+{
+    "a_galaxy_workflow": "true",
+    "annotation": "This workflow takes a collection and returns 2 collections. One with the items which contains the pattern, one with the items which do not contains the pattern.",
+    "comments": [],
+    "creator": [
+        {
+            "class": "Person",
+            "identifier": "https://orcid.org/0000-0002-1964-4960",
+            "name": "Lucille Delisle"
+        }
+    ],
+    "format-version": "0.1",
+    "license": "MIT",
+    "name": "Split collection by pattern in identifiers",
+    "report": {
+        "markdown": "\n# Workflow Execution Report\n\n## Workflow Inputs\n```galaxy\ninvocation_inputs()\n```\n\n## Workflow Outputs\n```galaxy\ninvocation_outputs()\n```\n\n## Workflow\n```galaxy\nworkflow_display()\n```\n"
+    },
+    "steps": {
+        "0": {
+            "annotation": "Collection you want to split into 2",
+            "content_id": null,
+            "errors": null,
+            "id": 0,
+            "input_connections": {},
+            "inputs": [
+                {
+                    "description": "Collection you want to split into 2",
+                    "name": "Input Dataset Collection"
+                }
+            ],
+            "label": "Input Dataset Collection",
+            "name": "Input dataset collection",
+            "outputs": [],
+            "position": {
+                "left": 0,
+                "top": 72
+            },
+            "tool_id": null,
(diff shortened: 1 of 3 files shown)


lldelisle:
@wm75 if you want to review this.

workflows/data-manipulation/split-collection/Split-collection-using-tabular-tests.yml
```diff
@@ -0,0 +1,56 @@
...
+        cat1_2:
+          asserts:
+            has_text:
+              text: "whatever"
+        cat1_3:
+          asserts:
+            has_text:
+              text: "whatever"
+
+
```
mvdbeek, line 56:
```suggestion
```

workflows/data-manipulation/split-collection/Split-collection-using-comma-separated-list-tests.yml
```diff
@@ -0,0 +1,52 @@
...
+              text: "whatever"
+        cat1_2:
+          asserts:
+            has_text:
+              text: "whatever"
+        cat1_3:
+          asserts:
+            has_text:
+              text: "whatever"
+
```
mvdbeek, line 52:
```suggestion
```

workflows/data-manipulation/split-collection/README.md
```diff
@@ -0,0 +1,14 @@
+# Split collection
+
+These workflows allow to split a collection into 2 using identifiers.
+
+The common input to all workflows is a collection of type 'list'.
+
+The way to split the collection differs with the workflow.
+
+- In the workflow "Split collection by pattern in identifiers", you need to specify a "pattern". This is a word that is present only in one part of your samples. This will split your collection into 2: one with the identifiers which have the 'pattern' and the other one with the identifiers which don't have.
```
mvdbeek, line 9: Is that a regex pattern or literal words ?

workflows/data-manipulation/split-collection/Split-collection-by-pattern-in-identifiers.ga
```diff
@@ -0,0 +1,244 @@
...
+        {
+            "class": "Person",
+            "identifier": "https://orcid.org/0000-0002-1964-4960",
+            "name": "Lucille Delisle"
+        }
+    ],
+    "format-version": "0.1",
+    "license": "MIT",
+    "release": "0.1",
+    "name": "Split collection by pattern in identifiers",
```
mvdbeek, line 15:
```suggestion
    "name": "Split collection by pattern in element identifiers",
```

mvdbeek (approved):

workflows/data-manipulation/split-collection/README.md
lldelisle: It uses grep so it is a regex. However, as this workflow is targeting a "user" audience I was trying to make it understandable for non bioinformaticians but I am happy to get suggestions.

lldelisle:
@mvdbeek, in the workflow: "Split-collection-using-tabular", I extract during the workflow the value of the group name I want to select. Is there a way to use this value to rename the final output collections?
I tried to define the json parameter as a workflow output and reuse the name of this output as `${my_variable}` but it just asked me its value when I tried to run the workflow which is not what I wanted...

workflows/data-manipulation/split-collection/README.md
mvdbeek: They might be surpised if they do enter special characters, so best to mention this I think ?

mvdbeek: Actually, do you expect users to run this workflow as a standalone workflow ? I thought this was useful as a thing to embed in another workflow ? Would it maybe make more sense to add what is not covered by the filter filter to the tool itself ?

lldelisle: I feel like I am using this workflow as a standalone. For example, to select only part of my samples for deeper analysis or for a specific heatmap etc...

You are right about the word/regex, I should warn the users.

PR #604, opened 2024-11-19
Credits Persist-seq original pipeline for the Pseudo-bulk scRNA-seq pipeline

pcm32:
Adds credits on the Pseudo-bulk scRNA-seq pipeline, and mentions main differences for user's awareness.

### DIFF
--- a/workflows/scRNAseq/pseudobulk-worflow-decoupler-edger/README.md
+++ b/workflows/scRNAseq/pseudobulk-worflow-decoupler-edger/README.md
@@ -4,6 +4,13 @@ This workflow uses the decoupler tool in Galaxy to generate pseudobulk counts fr
 using the edgeR tool. The workflow also includes data sanitation steps to ensure smooth operation of edgeR and minimizing potential issues. Additionally, a Volcano plot tool is used to visualize the results after the DEG 
 analysis.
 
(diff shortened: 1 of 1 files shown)


pavanvidem:
@pcm32 thanks! The workflows on Galaxy can still be changed after sharing/publication. I am not sure whether there is a way to freeze and version a workflow on Galaxy. If we are citing a specific version of the original workflow here, then it would be good to have this version frozen somewhere (like WorkflowHub or IWC). That's just an idea. Maybe there are better ways to do it.

pcm32:
Maybe we can add the file on a supplementary directory here if you prefer? I don't mind. I don't tend to unshare things off the main galaxy instances, but I understand why you suggest it.

pcm32:
Adding to IWC is probably too much work for what I'm after.

pavanvidem:
@mvdbeek is it allowed to have such a supplementary directory with an additional workflow? Is there a better solution to cite a specific version of a published Galaxy workflow? something similar to history achieving but for workflows?

mvdbeek:
I don't understand, all IWC workflows are versioned. I don't see a problem with the readme either.

mvdbeek (approved):

pavanvidem:
@mvdbeek, if you think we can keep that workflow link, please go ahead and merge.

mvdbeek:
It's a link, just like GTN or wikipedia links they can go away, I don't think this is a problem.

mvdbeek:

https://github.com/galaxyproject/iwc/actions/runs/11932563263

PR #613, opened 2024-11-21
add MIT licence - copied from tools-iuc

fubar2:


fubar2:
The galaxy code base is probably a better LICENCE document - updated

wm75:
Shouldn't this be accompanied with a clarification, which parts of the repo the license applies to, i.e. that the workflows themselves can specify distinct licenses?
Similar to what the GTN does in its github [README](https://github.com/galaxyproject/training-material/blob/main/README.md#license)?

mvdbeek:
The workflows have various licenses, and this one doesn't make sense, definitely nothing here is AFL.

fubar2:
Background: Trying to spray licenses to every galaxy project repo from a spreadsheet showing which ones are missing. The owners are the ones who can best choose the right one - I'm just trying to get the process of remediation started so any help in making it the right one would be greatly appreciated.

bgruening:
I adopted the README and added a MIT only license for all the rest. Let me know if this is ok.

mvdbeek (approved):
I don't love that it says MIT on the repo overview and that you will not notice the different licenses unless you look. The "LICENSE at the root" feature has gone from a handy feature that is optional to something people apparently use for evaluations.

bwlang:
I just did a quick test... it's possible to include a license.md file with text like this:
```
Various Licenses

Each tool in this repository specifies its own open source license.
```
that shows a license link like this:
image

It's kind of a bummer to overload this feature, but i think it's better than the other options:
1) incorrect license (very bad if it says one thing, but the workflow is more restrictive)
2) nothing

PR #620, opened 2024-11-30
First pass at generating mermaid diagrams for all workflows

mvdbeek:
These are pretty cool.
Screenshot 2024-12-09 at 15 44 29

 I guess some possible next steps might be
- [x] render the mermaid diagrams in the static site
- [x] render subworkflow steps into the same markdown file
- [ ] include diagram in readme.md file
- [ ] use hyperlinks to move users to subworkflows
- [ ] explore tool tips to maybe show tool ids

### DIFF
--- a/.github/workflows/gh-page-preview.yml
+++ b/.github/workflows/gh-page-preview.yml
@@ -13,8 +13,6 @@ jobs:
     steps:
       - name: Checkout PR Branch
         uses: actions/checkout@v4
-        with:
-          ref: ${{ github.head_ref }}
 
       - name: Set up Python
         uses: actions/setup-python@v5
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ tool_test_output.html
 tool_test_output.json
 .DS_Store
 workflow_manifest.json
+.vscode
--- a/scripts/create_mermaid.py
+++ b/scripts/create_mermaid.py
@@ -1,0 +1,97 @@
+import argparse
+import json
+import os
+from typing import Literal
+
+STEP_TYPE_TO_SHAPE = {
+    "data_input": "@{ shape: doc }",
+    "data_collection_input": "@{ shape: docs }",
+    "parameter_input": "@{ shape: lean-l }",
+    "tool": "@{ shape: process }",
+    "subworkflow": "@{ shape: subprocess }",
+}
+
+
+def step_to_mermaid_item(
+    step_type: Literal[
+        "parameter_input", "data_input", "data_collection_input", "tool", "subworkflow"
+    ],
+    step_label: str,
+):
+    step_label_anchor = f'["{step_label}"]'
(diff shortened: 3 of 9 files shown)


hexylena:
The corresponding GTN version: https://github.com/galaxyproject/training-material/blob/main/_plugins/jekyll-topic-filter.rb#L398-L457,

mvdbeek:
That's cool, I'll see that we can make that look more alike. The labels and utf-8 chars are nice!

hexylena:
I thought in the past about making it a service, e.g. on apps.galaxyproject.eu, where we could pass in a URL (or POST) a workflow file, and get back either the raw mermaid text (maybe good for your use case embedding in github), or an iframe that's ready to go (maybe good for the GTN use case.) Not sure if this is worth exploring but, wanted to raise it in case it was interesting to you as well.

mvdbeek:
I thought about that, but i think the most interesting use is in embedding the diagrams in markdown files and for that the tiny script seems like less moving parts ? We could definitely push this into tool-util for re-use though, and maybe use this as the svg rendering engine instead of the one that we have now in Galaxy (as used in workflow reports) ?

hexylena:
yeah, similar considerations for me, network requests are sub-optimal. I like the idea of being able to use it for reports, it's a nicer cleaner representation a lot of the time!

mvdbeek:
Turns out .ga is a lot easier to process than gxformat2. I spent the afternoon processing the gxformat2 workflows, and I think I haven't gotten anywhere. I could just convert yaml workflows to the native representation ...

mvdbeek:
I've gone the conversion to native route for now. There's a bunch more we could do but let's maybe say this is good enough for a first pass (and maybe some hacking next week) ?

bgruening:
Looks pretty cool! Is this replacing the embedded workflow version? Or in addition?

mvdbeek:
There's a lot of complexity to embedding the preview (need to add cors exceptions, need a handle to the workflow, need the tools installed etc) and I also prefer the simpler view here. I don't need to know what each step's input and output is.

mvdbeek:
Was the 👎 to technical issues (the way we run the IWC this isn't currently possible without committing to writing back workflow ids) or to me preferring the view without explicit input-output connections ? Is there something you would like to see ?

dannon:
Don't want to preclude more discussion about improvements or suggestions around any eventual/possible role of the live-galaxy embedded renderings, but the challenges Marius mentions and we've discussed are significant and this is fantastic stuff that would be great to go ahead and ship.  They work and look great in my local testing, so I'm going to go ahead and merge this.

dannon (approved):

mvdbeek:

https://github.com/galaxyproject/iwc/actions/runs/12268454675

jmchilton:
> I could just convert yaml workflows to the native representation ...

This is what all the other display stuff like the graph visualization does. I am sorry I didn't make that clearer.

PR #623, opened 2024-12-05
Fix stuck colormode in website

dannon:
If someone happened to visit iwc.galaxyproject.org with dark mode enabled prior to us swapping to a galaxy-themed light mode by default, the browser could get stuck in dark mode due to the way nuxt stores a key in local storage without unsetting it.  This is a workaround that unsticks the stuck mode.

FOR CONTRIBUTOR:

FOR REVIEWERS:


dannon:
We can ignore the failing build preview here -- that's also on my list for today to approach in a different way.

PR #631, opened 2024-12-11
Add workflow:  genome annotation with Maker

rlibouba:
Hello,
I would like to propose this workflow for annotating a genome  from the GTN "Genome annotation with Maker (short)".

Thank you! Have a nice day!


wm75:
This looks really cool already, thanks!

When I'm importing the WF on Galaxy Europe, it complains about

Screenshot From 2024-12-12 10-35-29

So I guess that needs to be fixed.

----

In the README I would not discuss inputs/outputs per tool but at the WF level to avoid redundancy.
You can just list the inputs, their expected format and what the WF will use them for.
Likewise, you can have one list of all outputs, what they are for and (if that's relevant/helpful) which tool produced them.

----

Without knowing too much about the details of the WF, would a MultiQC combined report of some of the results make sense?

workflows/genome_annotation/annotation_maker/Genome_annotation_with_maker_short_tests.yml
```diff
@@ -0,0 +1,104 @@
...
+      class: File
+      location: https://zenodo.org/records/13987091/files/augustus_training.tar.gz.augustus
+      filetype: augustus
+    SNAP training:
+      class: File
+      location: https://zenodo.org/records/13987091/files/snap_training.snaphmm
+      filetype: snaphmm
+  
+  
+  outputs:
```
wm75, line 25: Most of these outputs should be exactly reproducible text files. It's better to check them for actual content than just for approx. size.
You can test the presence of specific lines of text, for example.

rlibouba:
Hello, 
I have made some changes, in particular to the selection of the lineage for BUSCO. In this workflow, you run BUSCO twice using the same lineage. I'd like to do this so that the user can select the line only once : 
image

But I have a problem with the execution of BUSCO. For the lineage_dataset parameter instead of having `"--lineage_dataset fungi__odb10"`, I have this `"--lineage_dataset '__lt__galaxy.tools.parameters.workflow_utils.ConnectedValue object at 0x7fc7ac5a78d0__gt__'"`. 

I can't figure out what to put for this parameter. Does anyone have any ideas?

rlibouba:
Hi @mvdbeek, thank you for your help and recommendation for my problem. I have corrected the last errors.

workflows/genome_annotation/annotation-maker/Genome_annotation_with_maker_short.ga
```diff
@@ -177,7 +177,7 @@
...
..., \"poales_odb10\", \"polyporales_odb10\", \"primates_odb10\", \"propionibacteriales_odb10\", \"pseudomonadales_odb10\", \"pseudomonadota_odb10\", \"reptilia_odb10\", \"rhodobacterales_odb10\", \"rhizaria_odb10\", \"rhizobiales_odb10\", \"rhizobium-agrobacterium_group_odb10\", \"rhodospirillales_odb10\", \"rickettsiales_odb10\", \"saccharomycetes_odb10\", \"sauropsida_odb10\", \"selenomonadales_odb10\", \"solanales_odb10\", \"sphingobacteriia_odb10\", \"sphingomonadales_odb10\", \"spirochaetes_odb10\", \"spirochaetales_odb10\", \"spirochaetia_odb10\", \"streptomycetales_odb10\", \"streptosporangiales_odb10\", \"stramenopiles_odb10\", \"sulfolobales_odb10\", \"synechococcales_odb10\", \"synergistetes_odb10\", \"tardigrada_odb10\", \"tenericutes_odb10\", \"tetrapoda_odb10\", \"thaumarchaeota_odb10\", \"thermoanaerobacterales_odb10\", \"thermoplasmata_odb10\", \"thermoproteales_odb10\", \"thermoprotei_odb10\", \"thermotogae_odb10\", \"thiotrichales_odb10\", \"tissierellales_odb10\", \"tissierellia_odb10\", \"tremellomycetes_odb10\", \"vertebrata_odb10\", \"verrucomicrobia_odb10\", \"viridiplantae_odb10\", \"xanthomonadales_odb10\"], \"parameter_type\": \"text\", \"optional\": false}",
```
mvdbeek, line 180: Hmm, this is something we would have to update whenever the tool is being updates, what made you decide to undo the restrict values on connection option ?

rlibouba: It was to allow users to have a list with the different lineages. If they had to write the busco lineage, it would have to put “_odb10” at the end for the busco "lineage_dataset" parameter to work.

mvdbeek: I see now where this is going wrong, there's a bug with `restrictOnConnections` not populating the available options. This is supposed to be the exact same list you've written here.

rlibouba: Should we use `restrictOnConnections` ? How can correct this bug?

workflows/genome_annotation/annotation-maker/Genome_annotation_with_maker_short.ga
```diff
@@ -0,0 +1,913 @@
+{
+    "a_galaxy_workflow": "true",
+    "annotation": "",
```
mvdbeek, line 3: Can you put a short description here on what the workflow does ?
```suggestion
    "annotation": "This workflow allows for genome annotation using Maker and evaluates the quality of the annotation with BUSCO and genome annotation statistics. The annotation can then be improved, standardized, and visualized with additional tools.",
```

mvdbeek (approved):

PR #664, opened 2025-02-06
Use all tests

lldelisle:
I found 4 files 'tests' that were not cited in any dockstore:
```
for f in $(find . -name "*test*yml"); do n=$(grep -nr $(basename $f) workflows | wc -l); if [ $n = "0" ]; then echo $f; fi; done
```
I tried to solve all of them. I would like that authors of the workflow validate this PR:
- @RZ9082 
- @rlibouba 
- @pvanheus 
- @wm75 or @kciy


RZ9082:
Oh! Forgot to add this one! Thank you :)

PR #751, opened 2025-03-19
Changed folders

paulzierep:
This moves the foodborne-pathogen folder one level down, to allow for other microbiome workflows, and moves the amplicon-mgnify folder as mgnify in amplicon.
The workflow directories that are also in https://github.com/iwc-workflows are not touched!


mvdbeek:

https://github.com/galaxyproject/iwc/actions/runs/13955904413

PR #764, opened 2025-03-24
Update rnaseq-de-filtering-plotting.ga

pavanvidem:
Make header file tabular format to fix auto detection


mvdbeek (approved):
Could I also ask you to update https://github.com/bgruening/galaxytools/blob/master/tools/text_processing/text_processing/recurring_lines.xml#L23

with something like
```
    <sanitizer invalid_char="">
        <valid initial="default">
            <add value="&#9;" />  <!-- &#9; is the HTML entity for a tab -->
        </valid>
    </sanitizer>
```
so that the extra sed step can be removed ?

PR #768, opened 2025-03-25
Add annotation workflow with Braker3

rlibouba:
Hello,
I would like to propose this workflow for annotating a genome from the GTN "Genome annotation with Braker3 ".
Thank you! Have a nice day!
Romane

### DIFF
--- a/workflows/genome_annotation/annotation-braker3/CHANGELOG.md
+++ b/workflows/genome_annotation/annotation-braker3/CHANGELOG.md
@@ -1,0 +1,5 @@
+# Changelog
+
+## [0.1] - 2025-03-25
+
+Initial version of the genome annotation workflow with braker3.
--- a/workflows/genome_annotation/annotation-braker3/Genome_annotation_with_braker3.ga
+++ b/workflows/genome_annotation/annotation-braker3/Genome_annotation_with_braker3.ga
@@ -1,0 +1,660 @@
+{
+    "a_galaxy_workflow": "true",
+    "annotation": "This workflow uses Braker3 to annotate a genome.",
+    "comments": [
+        {
+            "child_steps": [
+                8
+            ],
+            "color": "pink",
+            "data": {
+                "title": "Visualization"
+            },
+            "id": 4,
+            "position": [
+                1040,
+                780
+            ],
+            "size": [
+                240,
+                244.8
+            ],
+            "type": "frame"
+        },
+        {
+            "child_steps": [
+                1,
+                3,
+                2
+            ],
+            "color": "yellow",
+            "data": {
+                "title": "Inputs"
+            },
+            "id": 0,
+            "position": [
+                0,
+                320
+            ],
+            "size": [
+                240,
+                402.2
+            ],
+            "type": "frame"
+        },
+        {
+            "child_steps": [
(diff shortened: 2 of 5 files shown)


mvdbeek:
You seem to have helixer changes in here, let me know if you want to remove those ?

abretaud:
@rlibouba you've seen that you still still have some helixer changes? ;)

workflows/genome_annotation/annotation-braker3/Genome_annotation_with_braker3.ga
```diff
@@ -0,0 +1,659 @@
+{
+    "a_galaxy_workflow": "true",
+    "annotation": "This workflow uses Braker3 to annotate a genome.",
```
mvdbeek, line 3: Could you extend this a little ? Maybe the first sentence from the readme ?]
```suggestion
    "annotation": "This workflow allows for genome annotation using Braker3 and evaluates the quality of the annotation with BUSCO and genome annotation statistics.",
```

Could you make the busco step optional ? I think there are different preferences wrt what to use to assess genome completeness but especially for large genomes we found busco to be a bottleneck.

mvdbeek:
> the list of lineages for the Busco lineage parameter is not complete.

There are 2 possibilities for that, the first is that some of the values you've entered manually are not valid and would fail if selected. The other is a bug, so it would be great if you can try running this on one of the values you only see if you list the options manually.


 > I wanted to make BUSCO analyses optional, but I was unable to do so. I don't know how to do it.

You toggle the "Conditionally skip steps" on the steps you want to skip like this:

Screenshot 2025-09-05 at 10 36 00



Then connect a boolean input parameter to the `when` parameter that appears:

Screenshot 2025-09-05 at 10 36 47


I also noticed this:

Screenshot 2025-09-05 at 10 40 10

Does it make sense to make the `cached_db` parameter configurable so you can use whatever is available on the galaxy instance you are using (the IWC tests use the cvmfs reference data that is also available at usegalaxy.org)

rlibouba:
Hello @mvdbeek and @Delphine-L , I am having a problem with BUSCO. Both analyses fail in the CI. 

There is a line that catches my attention: https://github.com/galaxyproject/iwc/actions/runs/17731976942/job/50385002292#step:7:3080

What is the CI launched on? I thought it was on EU. What I don't understand is that for another annotation workflow with Helixer, I had set the same database: `all+2024-03-21-114020`

mvdbeek:
On the github worker, with reference data from cvmfs, which is what usegalaxy.org uses.

rlibouba:
Hello @Delphine-L , Do you think these latest changes are good ?

rlibouba:
Hello @mvdbeek and @Delphine-L, do you think this PR can be merged? Or are there other changes that need to be made?
Have a nice day!

workflows/genome_annotation/annotation-braker3/CHANGELOG.md
```diff
@@ -0,0 +1,5 @@
+# Changelog
+
+## [0.1] - 2025-03-25
```
mvdbeek:
```suggestion
## [0.1] - 2025-12-05
```

PR #769, opened 2025-03-26
MAGs building workflow

paulzierep:
FOR CONTRIBUTOR:

I have difficulties writing tests for this workflow. Ideally I would like to write a test that checks the created bins, but bin IDs (so the name of the fasta files in dereplicated_genomes) are randomly assigned, so the names are different every time I run the workflow. Is it possible to check e.g. the number of elements on the collection, or some other loose test. I could not find any details for test docs.
If not, is it ok to check only the mutliQC report, if any of the workflow steps does not work, the report will change significantly and the test fail.


workflows/microbiome/mags-building/MAGs-generation.ga
```diff
@@ -0,0 +1,2630 @@
+{
+    "a_galaxy_workflow": "true",
+    "annotation": "This workflow constructs Metagenome-Assembled Genomes (MAGs) using SPAdes or MEGAHIT as assemblers, followed by binning with four different tools and refinement using Binette. The resulting MAGs are dereplicated across the entire input sample set, then annotated and evaluated for quality.\nYou can provide pooled reads (for co-assembly/binning), individual read sets, or a combination of both. The input samples must consist of the original reads, which are used for abundance estimation. In all cases, reads should be trimmed, adapters removed, and cleaned of host or other contaminants before processing.\n\nAttention: this workflow attempts to choose the required databases for the tools based on the availability on the server, please carefully check that the DBs you want to use are correctly chosen - in doubt use the latest versions available !\n",
```
mvdbeek, line 3:
```suggestion
    "annotation": "This workflow constructs Metagenome-Assembled Genomes (MAGs) using SPAdes or MEGAHIT as assemblers, followed by binning with four different tools and refinement using Binette. The resulting MAGs are dereplicated across the entire input sample set, then annotated and evaluated for quality.\nYou can provide pooled reads (for co-assembly/binning), individual read sets, or a combination of both. The input samples must consist of the original reads, which are used for abundance estimation. In all cases, reads should be trimmed, adapters removed, and cleaned of host or other contaminants before processing.",
```
that's not actionable to users, and if you were serious about this, would have to be listed on every tool parameter consuming anything from a tool data table.

If you want to provide further instructions I would place that into the relevant parameter (is this about the checkM2 tables) ?

workflows/microbiome/mags-building/MAGs-generation-tests.yml
```diff
@@ -0,0 +1,69 @@
...
+    Custom Assemblies: null
+    Minimum length of contigs to output: '200'
+    Read length (CONCOCT): '100'
+    Environment for the built-in model (SemiBin): global
+    Contamination weight (Binette): '2'
+    CheckM2 Database for Binette: 1.0.2
+    Minimum MAG completeness percentage: '1'
+    Maximum MAG contamination percentage: '25'
+    Minimum MAG length: '100'
+    CheckM2 Database (dRep step): 1.0.2
```
mvdbeek, line 41: Why do these need to be set sepatately ?

workflows/microbiome/mags-building/MAGs-generation-tests.yml
```diff
@@ -0,0 +1,69 @@
+- doc: Test for Metagenome-Assembled-Genomes-(MAGs)-generation
+  job:
+    Trimmed grouped paired reads:
```
mvdbeek, line 3:
```suggestion
    Trimmed grouped:
```
that's redundant with what's shown in the workflow form.

workflows/microbiome/mags-building/MAGs-generation.ga
```diff
@@ -0,0 +1,2630 @@
...
+        "5": {
+            "annotation": "Provide already quality controlled trimmed reads, this can also be read groups or groups and individual samples or a mix of both.",
+            "content_id": null,
+            "errors": null,
+            "id": 5,
+            "input_connections": {},
+            "inputs": [
+                {
+                    "description": "Provide already quality controlled trimmed reads, this can also be read groups or groups and individual samples or a mix of both.",
+                    "name": "Trimmed grouped paired reads"
```
mvdbeek, line 184:
```suggestion
                    "name": "Trimmed reads"
```

workflows/microbiome/mags-building/MAGs-generation.ga
```diff
@@ -0,0 +1,2630 @@
...
+                }
+            ],
+            "label": "Trimmed grouped paired reads",
```
mvdbeek, line 187:
```suggestion
            "label": "Trimmed reads",
```

workflows/microbiome/mags-building/README.md
```diff
@@ -0,0 +1,32 @@
...
+- **Quality Control** via QUAST and CheckM/CheckM2  
+- **Abundance Estimation** per sample with CoverM  
+
+All results are consolidated into a single **MultiQC report** for easy analysis.  
+
+## Input Requirements  
+
+- **Quality-Controlled Reads**: Input reads must be quality-filtered, with host reads removed.  
+- **Trimmed Sample Paired Reads**: Quality-trimmed reads from individual samples, used solely for abundance estimation.  
+- **Trimmed Grouped Paired Reads**: These can be grouped based on the desired MAGs generation approach:
```
mvdbeek, line 26: Can you explain how these are different from Quality-Controlled Reads ?

workflows/microbiome/mags-building/README.md
```diff
@@ -0,0 +1,32 @@
...
+- **Abundance Estimation** per sample with CoverM  
+
+All results are consolidated into a single **MultiQC report** for easy analysis.  
+
+## Input Requirements  
+
+- **Quality-Controlled Reads**: Input reads must be quality-filtered, with host reads removed.  
+- **Trimmed Sample Paired Reads**: Quality-trimmed reads from individual samples, used solely for abundance estimation.  
+- **Trimmed Grouped Paired Reads**: These can be grouped based on the desired MAGs generation approach:  
+  - **Individual MAGs Generation**: Use the same input as `Trimmed Sample Paired Reads` to generate MAGs per sample.
```
mvdbeek, line 27: I don't understand why you have the user supply the same input twice ?

workflows/microbiome/mags-building/MAGs-generation.ga
paulzierep: the problem is, that I cannot specify a default and the initially provided DB is not the newest one (e.g. on EU), I want to prevent the user from using an old DB ... not sure how to communicate this to the user in a better was, any suggestion ?

workflows/microbiome/mags-building/MAGs-generation-tests.yml
paulzierep: I don't understand. This input are reads that are grouped. This allows for co-assembly. The workflow for grouping will also be submitted soon. If paired it not needed I would go for:
```suggestion
    Grouped trimmed reads:
```

workflows/microbiome/mags-building/MAGs-generation.ga
paulzierep:
```suggestion
                    "name": " Grouped trimmed reads:"
```

workflows/microbiome/mags-building/MAGs-generation.ga
paulzierep:
```suggestion
            "label": "Grouped trimmed reads",
```

workflows/microbiome/mags-building/README.md
paulzierep: The reads can be grouped. E.g. by grouping all reads into one file one performs co-assembly/binning Grouping in groups one performs grouped-assembly/binning

paulzierep:
Hi @mvdbeek, it seems the latest test was successful: https://github.com/galaxyproject/iwc/actions/runs/14382300139?pr=769 
Was the explanation for Trimmed reads / Grouped trimmed reads OK ? Can we do anything else to get this further ?

workflows/microbiome/mags-building/MAGs-generation.ga
```diff
@@ -0,0 +1,2576 @@
+{
+    "a_galaxy_workflow": "true",
+    "annotation": "This workflow constructs Metagenome-Assembled Genomes (MAGs) using SPAdes or MEGAHIT as assemblers, followed by binning with four different tools and refinement using Binette. The resulting MAGs are dereplicated across the entire input sample set, then annotated and evaluated for quality.\nYou can provide pooled reads (for co-assembly/binning), individual read sets, or a combination of both. The input samples must consist of the original reads, which are used for abundance estimation. In all cases, reads should be trimmed, adapters removed, and cleaned of host or other contaminants before processing.",
+    "comments": [],
+    "creator": [
+        {
+            "class": "Person",
+            "identifier": "https://orcid.org/0000-0001-9852-1987",
+            "name": "B\u00e9r\u00e9nice Batut"
```
mvdbeek, line 9:
```suggestion
            "name": "Bérénice Batut"
```

mvdbeek (approved):
Awesome, thank you!

workflows/microbiome/mags-building/README.md
```diff
@@ -0,0 +1,33 @@
...
+- **Annotation** with Bakta  
+- **Taxonomic Assignment** using GTDB-Tk  
+- **Quality Control** via QUAST and CheckM/CheckM2  
+- **Abundance Estimation** per sample with CoverM  
+
+All results are consolidated into a single **MultiQC report** for easy analysis.  
+
+## Input Requirements  
+
+Input reads must be quality-filtered, with host reads removed.
```
mvdbeek, line 24: I'm ok with the current state as the first version, but I opened the workflow run form for the workflow and I have a hard time understanding what I would need to do. I think in part that can be improved by adding a bit of workflow logic instead of requiring users to provide collections of varying shapes ?

Screenshot 2025-04-28 at 15 12 40
That should ideally be a conditional setting where you only provide the input if custom assembly is chosen. That's not possible yet, but I think if we had conditionals in the run form this would make the workflow a little easier to use, though I can understand this part.

I struggle more with this:
Screenshot 2025-04-28 at 15 14 20
Is this what you meant by:

> The reads can be grouped. E.g. by grouping all reads into one file one performs co-assembly/binning Grouping in groups one performs grouped-assembly/binning

Could you instead add a boolean "Perform grouped assembly?" and if the boolean is  true you use the original input collection and if it is false you merge the reads ? That way you would only have one input and it would be relatively clear to the user what inputs they need to provide ?

paulzierep: `That should ideally be a conditional setting where you only provide the input if custom assembly is chosen.`
I completely agree that would be better - if we had this option.
`Could you instead add a boolean "Perform grouped assembly?" and if the boolean is true you use the original input collection and if it is false you merge the reads ? That way you would only have one input and it would be relatively clear to the user what inputs they need to provide ?`
Perform grouped assembly - will require another workflow to be run, since it can also require a table that defines the grouping, here again I think the workflow logic does not allow to put it all in one workflow. This workflow will also be added here soon.

mvdbeek: ok, we can revisit that when you have the workflow ready.

paulzierep:
what needs to be done to finish: `Check workflow success Expected — Waiting for status to be reported` -sorry for the rush, it would be great to have this merged for project reporting

PR #818, opened 2025-04-28
add align, chain, and net workflow

martenson:
FOR CONTRIBUTOR:
* [ ] I have read the [Adding workflows guidelines](https://github.com/galaxyproject/iwc/blob/main/workflows/README.md#adding-workflows)
* [ ] License permits unrestricted use (educational + commercial)
* [ ] Please also take note of the reviewer guidelines below to facilitate a smooth review process.

FOR REVIEWERS:


martenson:
@mvdbeek I am not sure if there is some non-gpu hack, I'll have a closer look. The test runs fine against vgp.usegalaxy.org but that is likely not very helpful here.

workflows/genome-alignment/align-chain-net/align.ga
```diff
@@ -0,0 +1,921 @@
...
+    "creator": [
+        {
+            "class": "Organization",
+            "name": "Penn State University",
+            "url": "https://www.ccbb.psu.edu/"
+        },
+        {
+            "class": "Person",
+            "identifier": "https://orcid.org/0000-0002-9318-1781",
+            "name": "Martin \u010cech"
```
martenson, line 131: @mvdbeek 👀

martenson: I am just surprised they have different capitalizations of `C`

mvdbeek:
I think we'll have to invent some way to say that a workflow needs to run against a live instance when it really can't run in CI. Or perhaps the better way is to setup a "only for IWC" gpu pulsar destination. Then we only launch the step that needs special stuff against the remote resource.

PR #929, opened 2025-08-08
Add genome assembly QC post assembly workflow

hugolefeuvre:
Hello,

The purpose of this PR is to add a new workflow that allows quality and contamination control based on contigs once assembly is complete. This workflow is the source of the changes made in this PR concerning QC based on raw reads (https://github.com/galaxyproject/iwc/pull/927).

Thanks 😃

FOR CONTRIBUTOR:

FOR REVIEWERS:

### DIFF
--- a/workflows/genome-assembly/quality-and-contamination-control-post-assembly/.dockstore.yml
+++ b/workflows/genome-assembly/quality-and-contamination-control-post-assembly/.dockstore.yml
@@ -1,0 +1,19 @@
+version: 1.2
+workflows:
+- name: main
+  subclass: Galaxy
+  publish: true
+  primaryDescriptorPath: /quality_and_contamination_control_post_assembly.ga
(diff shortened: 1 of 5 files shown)


hugolefeuvre:
Hi @mvdbeek, if it's ok with you, we'd like to know if it's possible to merge this WF and the Raw Reads QC one (https://github.com/galaxyproject/iwc/pull/927). We know that we have to improve them with collection but as I said, we are currently working on it and we hope to add it soon. 
Right now what we'd like to do is to create 2 meta-WF with WFs that we implemented in IWC, one fastQ meta-WF from raw reads QC to bacterial genome annotation, and one fasta meta-WF from fasta QC to bacterial genome annotation. But for that, we need the latest versions of the WFs, which is why I am following up with you.

mvdbeek:
Thank you!

PR #1033, opened 2025-12-03
Enhance README with built-in index usage details

paulzierep:
Since this was not clear to many workflow creators and I had to point to the right locations, I thought we could add it here.


mvdbeek:
Thank you!

PR #1053, opened 2025-12-16
Add ecology "pool of amplicons map to reference" workflow

mo-mahe:
Workflow for processing amplicon pool sequencing data with reference.

This workflow allows you to reconstruct a sequence from an amplicon pool using a reference sequence. To run this workflow, you need the reads from the pool library you want to analyse in FASTQ format, separated into two files: forward and reverse. You will also need your reference sequence in FASTA format. This workflow creates a consensus sequence and a metadata file containing the length of the consensus sequence, the number of reads mapped to it, and the average, minimum, and maximum coverage depth. You can also retrieve a file containing unmapped reads.


mo-mahe:
@yvanlebras

PR #1062, opened 2026-01-10
Add TB variant analysis workflow

pvanheus:
FOR CONTRIBUTOR:

FOR REVIEWERS:

### DIFF
--- a/workflows/bacterial_genomics/tb-variant-analysis/.dockstore.yml
+++ b/workflows/bacterial_genomics/tb-variant-analysis/.dockstore.yml
@@ -1,4 +1,4 @@
-version: 1.5.2
+version: 1.2
 workflows:
 - name: main
   subclass: Galaxy


mvdbeek:
```
Failed to run bamqc
java.lang.RuntimeException: The BAM file is empty or corrupt
```

it seems that snippy produced a BAM file without alignments, did you expect that ?

pvanheus:
The error is coming from the fact that the kraken2 database essential for this workflow is not available in the test environment. Instead `/cvmfs/data.galaxyproject.org/managed/kraken2_databases/2020-11-24T164216Z_silva_kmer-len_35_minimizer-len_31_minimizer-spaces_6_load-factor_0.7` gets used, no reads get the correct classification and everything goes wrong from there. The workflow needs `2024-10-12T100500Z_Mycobacterium_v1_doi_10.5281_zenodo.8339822` which is available on usegalaxy.eu, usegalaxy.org and usegalaxy.org.au.

I started on a version of the `kraken2` wrapper that would allow the database to be a (Directory) parameter here: https://github.com/galaxyproject/tools-iuc/pull/7257. If "works on all Galaxy servers" is a requirement, then it will be necessary to complete the work start in that PR before merging (a future version of) this PR.

mvdbeek:
We mount the same cvmfs that usegalaxy.org mounts, are you sure that's the right diagnosis ?

pvanheus:
@mvdbeek this is what I currently see as the contents of that directory:

image

Whereas when I run `kraken2` with the Mycobacterial database on .eu the path used is `/data/db/data_managers/kraken2_databases/Mycobacterium_v1_doi_10.5281_zenodo.8339822_from_v1_assembly_summary.txt_kmer-len_35_minimizer-len_31_minimizer-spaces_7_load-factor_0.7` - i.e. not on CVMFS. I had a look, and I was wrong - the Mycobacterial database is only available on .eu, not .org.au or .org. So: one way to address the current problem would be to include the Mycobaterial database in CVMFS.

PR #1066, opened 2026-01-16
Add Taxonomy Assignment with QIIME2 Workflow

tflowers15:
FOR CONTRIBUTOR:

FOR REVIEWERS:


tflowers15:
No problem. Thank you for the clarification on what types of workflows are and are not suitable for the IWC. I will move to publish this workflow on WorkflowHub. I must have overlooked that WorkflowHub published workflows are given a DOI.

On the points that you had identified as being useful additions, I think adding the biom output and creating a workflow to train a classifier would be straight forward. In order to choose alternate classifiers (e.g. bayes vs sklearn) in workflow III, would the solution be to create a second version of the workflow (e.g. workflow IIIa and IIIb, similar to workflow IIa and IIb for single vs paired end data) using the alternate classifier?

For adding a biom output, that would be running `qiime2 tools export` on the output feature-table and updating the workflows IIa and IIb to include this step?
- "QIIME2 IIa: Denoising (sequence quality control) and feature table creation (single-end)"
- "QIIME2 IIb: Denoising (sequence quality control) and feature table creation (paired-end)"

What is the standard method for updating existing workflow? Is it to import the workflow into a `usegalaxy.*`, update the workflow and workflow tests and then create a pull request with the `.ga` and `-tests.yaml` files replaced in the workflow's directory?

The workflow to train the classifier would take the reference database `fasta` sequences and `tabular` taxonomy table as inputs and produce an output `classifier` which can then be used as an input for the "QIIME2-III-V-Phylogeny-Rarefaction-Taxonomic-Analysis" workflow? This would be a new workflow submission.

PR #1162, opened 2026-03-19
Update fasttree 2.1.10+galaxy1 to 2.1.10+galaxy3 in PathoGFAIR workflow.

jmchilton:
Both intermediate releases are Galaxy-level bug fixes with no input/output changes:
- galaxy1→galaxy2: fix version_command (galaxyproject/tools-iuc#3114)
- galaxy2→galaxy3: fix tool XML declaration (galaxyproject/tools-iuc#6640)

The toolshed doesn't have the version listed in the workflow anymore - looking over galaxyproject/tools-iuc#6640 I can imagine why. My first thought was maybe we broke the tool loading in https://github.com/galaxyproject/galaxy/pull/15493. But that behavior was trying to block on a tool version. The problem isn't immediately clear but I could put more time into if want to push back and call this a tool shed bug - I could try to re-implement boolean parameters that have matching truevalues and falsevalues if they are empty. 

FOR CONTRIBUTOR:

FOR REVIEWERS:

### DIFF
--- a/workflows/microbiome/pathogen-identification/pathogen-detection-pathogfair-samples-aggregation-and-visualisation/CHANGELOG.md
+++ b/workflows/microbiome/pathogen-identification/pathogen-detection-pathogfair-samples-aggregation-and-visualisation/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## [0.1.1] - 2026-03-19
+
+### Changed
+- Update fasttree from 2.1.10+galaxy1 to 2.1.10+galaxy3
+
 ## [0.1] - 2024-04-24
 
 First release.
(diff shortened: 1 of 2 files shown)


jmchilton:
> [!NOTE]
> Posted by Claude (AI assistant) on behalf of @jmchilton — not authored by them personally.

Superseded by #1293, which carries this fasttree bump forward onto current `main` and adds the test-schema regularization for this workflow (split out of #1286).

Two reasons for a fresh PR rather than a rebase:

1. **This PR was incomplete.** It updated `content_id`, `tool_id`, and `changeset_revision` on the fasttree steps, but left `tool_version` at `2.1.10+galaxy1` in all three. #1293 fixes that.
2. **The red here is unrecoverable.** `Combine chunked test results` failed, which means a test failed — the `Test workflows` job reports success even when tests inside it fail, so its green tick was misleading. The artifacts and job logs from the 2026-03-19 run have since expired (`410 Gone`), so there is no way to recover *which* test failed.

Worth flagging for whoever picks up #1293: the fasttree pin is what stops PathoGFAIR from being invoked at all, so its workflow test has effectively never run on CI. Fixing the pin lets it execute for the first time, which is likely why this PR went red back in March — it probably surfaced a downstream failure that is normally masked. #1293 will give a fresh, readable signal.

The pin itself is confirmed still broken: the Toolshed's only installable fasttree revisions today are `2.1.10` and `2.1.10+galaxy3`, and PathoGFAIR errors identically on `main` in the weekly runs of 2026-07-06 and 2026-07-13.

PR #1177, opened 2026-03-23
Add workflow Metadata and Sequences from BioProjectIDs

gdefazio:
FOR CONTRIBUTOR:

FOR REVIEWERS:


gdefazio:
Hi @mvdbeek, I tried to integrate `parallel-accession-download-main` but I had lot of problems with scheduling of job for apply-rule tool of this wf because of too nested structure. Then I integrated fasterq-dump in my wf and I think that thanks to your suggestion now the wf is better. I'm waiting for CI start from yesterday but it seems stacked.

lldelisle:
Hi @gdefazio ,
I am currently running the workflow to really understand what it is doing. I will give my comments before one week.

lldelisle:
Hi @gdefazio,
The workflow takes as input a list of project IDs, then it generates one file per project ID with all SRA inside. This is given to fasterq-dump and then the collections are flatten to get a `list` collection for SR and a `list:pair` collection for PE.

I think your workflow is a great idea but I have few problems with the current workflow:
- The number of SRA per project can be large and therefore the fasterqdump step can be very long and I agree with @mvdbeek it would be better to use the subworkflow "Parallel accession download".
- To get the SRA the workflow takes the first column however, I feel like if you do not choose `detailed (metadata download) ` to `true` then `accession_run` is present in the file but is not the first column.

I have noticed something that could be improved:
- The identifiers of the metadata tables should be the project IDs

So I propose here a new version: https://usegalaxy.eu/u/delislel/w/metadata-and-sequences-from-bioproject-ids-final-ld

What I did compared to your workflow:
- I used the Parallel accession download as subworkflow and as input I simply concatenated all the files with SRA into one (you loose the info of from which project it comes from but in your version as at the end you flatten the collections I would say it is the same).
- I used a awk step to find the column ID with 'run_accession' and output the SRA list in case the user has not put `detailed (metadata download)` to `true`.
- I inserted a step to relabel the metadata tables with the project ID.

I let you try it and tell me what you think.

@mvdbeek is there a policy in case a workflow uses another workflow as subworkflow? Do they need to be in the same directory?

Don't hesitate if you have questions or remarks, I would be happy to answer.

mvdbeek:
Subworkflows are just embedded inside the parent. We're working on making those references too (https://github.com/galaxyproject/galaxy/pull/21887) and use symlinks in the IWC but for now this is fine as is and will just work.

gdefazio:
Hi @lldelisle and thank you for the valuable work on this wf and for positive feedback.
Let me answer you by poiny.

> * The number of SRA per project can be large and therefore the fasterqdump step can be very long and I agree with @mvdbeek it would be better to use the subworkflow "Parallel accession download".

After Marcus comment I tried to integrate "Parallel accession download" in this WF but I had some problems with the "apply rules" step because of issue with nested structure non matching what expected. Thank you for solving that.

> * To get the SRA the workflow takes the first column however, I feel like if you do not choose `detailed (metadata download) ` to `true` then `accession_run` is present in the file but is not the first column.

I was not aware about that, thanks for solving it.

> I have noticed something that could be improved:
> 
> * The identifiers of the metadata tables should be the project IDs
> 
> So I propose here a new version: https://usegalaxy.eu/u/delislel/w/metadata-and-sequences-from-bioproject-ids-final-ld
> 
> What I did compared to your workflow:
> 
> * I used the Parallel accession download as subworkflow and as input I simply concatenated all the files with SRA into one (you loose the info of from which project it comes from but in your version as at the end you flatten the collections I would say it is the same).

Do you suggest to have one fastq collection for each BioProject ID?

> * I used a awk step to find the column ID with 'run_accession' and output the SRA list in case the user has not put `detailed (metadata download)` to `true`.
> * I inserted a step to relabel the metadata tables with the project ID.
> 
> I let you try it and tell me what you think.

I tried it and I think is so much better than my version. Thank you.
> 
> @mvdbeek is there a policy in case a workflow uses another workflow as subworkflow? Do they need to be in the same directory?
> 
> Don't hesitate if you have questions or remarks, I would be happy to answer.

Can I add you as WF author?

Thanks again for your effort.

gdefazio:
Hi @lldelisle please give me some feedbacks to the previous message when you have time. Thanks in advance

lldelisle:
Hi,
Yes I would be happy to be an author.
I don't think it is necessary to return one collection per project.

workflows/data-fetching/metadata-and-sequences-from-BioProjectIDs/metadata-and-sequences-from-bioproject-ids-tests.yml
```diff
@@ -0,0 +1,76 @@
...
+    BioProject IDs:
+      class: File
+      path: test-data/test1_single_prj_pe.txt
+      filetype: txt
+    assay (metadata download): false
+    desc (metadata download): false
+    detailed (metadata download): true
+    expand (metadata download): false
+    Group by Experiments (fastq download): false
+    Group by Sample (fastq download): false
```
lldelisle: I think these parameters disappeared from the workflow

workflows/data-fetching/metadata-and-sequences-from-BioProjectIDs/metadata-and-sequences-from-bioproject-ids-tests.yml
```diff
@@ -0,0 +1,76 @@
...
+    BioProject IDs:
+      class: File
+      path: test-data/test2_multiple_prj_mixed.txt
+      filetype: txt
+    assay (metadata download): false
+    desc (metadata download): false
+    detailed (metadata download): true
+    expand (metadata download): false
+    Group by Experiments (fastq download): false
+    Group by Sample (fastq download): false
```
lldelisle: I think these parameters disappeared from the workflow

workflows/data-fetching/metadata-and-sequences-from-BioProjectIDs/metadata-and-sequences-from-bioproject-ids-tests.yml
```diff
@@ -0,0 +1,76 @@
...
+- doc: Test 2 for Metadata-and-Sequences-from-BioProjectIDs
+  job:
+    BioProject IDs:
+      class: File
+      path: test-data/test2_multiple_prj_mixed.txt
+      filetype: txt
+    assay (metadata download): false
+    desc (metadata download): false
+    detailed (metadata download): true
+    expand (metadata download): false
```
lldelisle: I would use a different combination from the first test, to increase the test coverage.

workflows/data-fetching/metadata-and-sequences-from-bioproject-ids/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+````
+PRJNA1425250
+PRJNA1417619
+PRJNA1425251
+PRJNA1417617
+PRJNA1425252
+PRJEB1417616
+````
+
```
lldelisle, line 18: Would you mind to describe here the parameters you expose, to guide the users?

workflows/data-fetching/metadata-and-sequences-from-BioProjectIDs/metadata-and-sequences-from-bioproject-ids.ga
```diff
@@ -0,0 +1,808 @@
...
+            "class": "Person",
+            "identifier": "https://orcid.org/0000-0003-2982-388X",
+            "name": "Paul Zierep"
+        }
+
+    ],
+    "format-version": "0.1",
+    "license": "CC-BY-4.0",
+    "release": "0.1",
+    "name": "Metadata and Sequences from BioProject IDs FINAL (imported from URL)",
```
lldelisle:
```suggestion
    "name": "Metadata and Sequences from BioProject IDs",
```

workflows/data-fetching/metadata-and-sequences-from-BioProjectIDs/metadata-and-sequences-from-bioproject-ids.ga
```diff
@@ -0,0 +1,808 @@
...
+    "readme": "# Metadata and Sequences from BioProject IDs\n\n## Rationale\nIn order to promote re-analysis of publicly available sequences from INSDC databases, we propose *Metadata and Sequences from BioProject IDs* a Galaxy workflow that starting by a list of valid BioProject IDs (e.g. PRJNA....., PRJEB.....) is able to manage data and metadata download.\n\n## Usage\nUpload a text file in which there is a BioProject ID for each row and run the workflow.\n\nThere is also the possibility to set optional options to regulate behaviour of metadata and data download.\n",
```
lldelisle: I would put the same readme in the workflow as in the repository

lldelisle (commented):
Thanks for updating the workflows and tests, here are my new comments.

workflows/data-fetching/metadata-and-sequences-from-BioProjectIDs/metadata-and-sequences-from-bioproject-ids-tests.yml
gdefazio: We have 4 options for this WF. These are not specific for the wf but all come from pysradb_search tool. I suggests to add tests in IUC pysradb_search tool repo if a broader tests coverage is needed.

lldelisle: The goal is not here to test the tool but to test that the workflow works with at least 2 different sets of combination. This is when I played with them that I discovered that if you put `detailed` to false the sra ID is not in the first column, so I think it worth to use another combination.

gdefazio: Ok, I see your point. I added another test similar to test 1 with different combination of parameters.

lldelisle (approved):
Looks good to me but because I contributed, I would prefer @mvdbeek to have a look and merge.

lldelisle:
@mvdbeek is it OK for you?

mvdbeek:
Thanks for the workflow! I reviewed it against the IWC checklist and pushed a commit (`73be6f9`) addressing a few naming/metadata consistency items:

- **Folder renamed** to `metadata-and-sequences-from-bioproject-ids` (all lowercase, dashes) to match IWC folder conventions — the folder becomes the repo name in `iwc-workflows` and part of the TRS ID. This also aligns it with the `.ga` filename.
- **`.dockstore.yml`** regenerated with `planemo dockstore_init` so it includes all three workflow creators (Giuseppe Defazio, Lucille Delisle, Paul Zierep) with ORCIDs — previously only the first author was listed.
- **`.workflowhub.yml`** slug typo fixed: `bioprojects-ids` → `bioproject-ids`.
- **Test `doc:` names** aligned with the human-readable workflow name.

The three canonical spellings (folder, `.ga` filename, WorkflowHub slug) now agree. The workflow file itself was untouched. The rest of the checklist looks good — human-readable input/output labels matching the tests file, a compliant annotation, changelog entry, and small test-data files (no Zenodo needed).

mvdbeek (approved):

PR #1188, opened 2026-03-27
Add RNA-seq Variant Discovery (RNAvar) workflow

emregulben:
## Summary
Adds the RNA-seq Variant Discovery (RNAvar) workflow.

## Technical Details
Implementation based on `nf-core/rnavar` logic:
* **Alignment:** STAR (2-pass mode).
* **Preprocessing:** GATK4 `MarkDuplicates`, `SplitNCigarReads`, and `BaseRecalibrator`.
* **Variant Calling:** GATK4 `HaplotypeCaller` using RNA-seq parameters.
* **Annotation:** SnpEff (hg38).

## Validation
* **Tests:** Planemo-validated suite verifying workflow outputs and technical parity with reference results.

FOR CONTRIBUTOR:

FOR REVIEWERS:


lldelisle:
If I may add to copilot review, I think would be good to describe in the README each of the expected inputs and the expected outputs.
Also add the README in the workflow so users can display it as 'help for this workflow'.
I am not familiar with the field, do you usually provide a fasta or do you use built-in genome indexes like hg38 etc..?

workflows/variant-calling/rna-variant-discovery/README.md
```diff
@@ -0,0 +1,14 @@
...
+
+This workflow is an implementation of a standard RNA-seq variant discovery pipeline. It handles the transition from raw reads to annotated variants using industry-standard tools.
+
+## Workflow Logic
+1. **Alignment:** **STAR** in 2-pass mode for high-accuracy splice-aware mapping.
+2. **Preprocessing:** **MarkDuplicates** and **GATK4 SplitNCigarReads** to prepare the BAM for variant calling by splitting reads into exon segments.
+3. **Recalibration:** **GATK4 BaseRecalibrator** (BQSR) using known polymorphic sites to adjust base quality scores.
+4. **Variant Calling:** **GATK4 HaplotypeCaller** with parameters specifically tuned for RNA-seq (e.g., ignoring soft-clipped bases).
+5. **Filtering:** **bcftools filter** applying hard filters ($FS > 30.0$, $QD < 2.0$) to reduce false positives.
+6. **Annotation:** Functional annotation of variants using **SnpEff** against the hg38 database.
```
wm75: It makes limited sense to hard-code the SnpEff database and, thereby, restrict analysis to just one organism and genome version. The database choice should instead be turned into a workflow parameter.

I would populate the user choices from the "Locally installed SnpEff databases" instead of letting the user download databases on demand.

Please let us know if you need help with this.

workflows/variant-calling/rna-variant-discovery/README.md
```diff
@@ -0,0 +1,14 @@
+# RNA-seq Variant Discovery (RNAvar)
+
+This workflow is an implementation of a standard RNA-seq variant discovery pipeline. It handles the transition from raw reads to annotated variants using industry-standard tools.
+
```
wm75, line 4: Here, the README should clearly state that the aim is to provide a faithful reimplementation of https://nf-co.re/rnavar and that currently there are still limitations.

workflows/variant-calling/rna-variant-discovery/README.md
```diff
@@ -0,0 +1,14 @@
...
+## Workflow Logic
+1. **Alignment:** **STAR** in 2-pass mode for high-accuracy splice-aware mapping.
+2. **Preprocessing:** **MarkDuplicates** and **GATK4 SplitNCigarReads** to prepare the BAM for variant calling by splitting reads into exon segments.
+3. **Recalibration:** **GATK4 BaseRecalibrator** (BQSR) using known polymorphic sites to adjust base quality scores.
+4. **Variant Calling:** **GATK4 HaplotypeCaller** with parameters specifically tuned for RNA-seq (e.g., ignoring soft-clipped bases).
+5. **Filtering:** **bcftools filter** applying hard filters ($FS > 30.0$, $QD < 2.0$) to reduce false positives.
+6. **Annotation:** Functional annotation of variants using **SnpEff** against the hg38 database.
+
```
wm75, line 12: After presenting the current workflow logic, it wouldbe good to have a "Limitations" section, where you discuss which features of the nf-core ppeline are currently not yet implemented.

Things hat I can immediately think of here are
- *Sample Sheet* support
- support for SnpEff alternatives for variant annotation
- optional UMI-Tools-based preprocessing
- no parallelization of the GATK4 SplitNCigarReads step (is this something that should be implemented at the tool level similar to what's done in the [freebayes wrapper](https://github.com/galaxyproject/tools-iuc/blob/0aeeab33a148d731eb4b2684de1d450c5b3e4bb0/tools/freebayes/freebayes.xml#L49-L62)?
- base recalibration and variant filtering are currently non-optional when they are optional in th nf-core pipeline

Basically, compare https://nf-co.re/rnavar/1.2.3/docs/usage/ to what's implemented here.
Of course, if you still want to fix any of these limitations, you're welcome to do so.

workflows/variant-calling/rna-variant-discovery/README.md
```diff
@@ -0,0 +1,14 @@
...
+## Testing and Parity
+The workflow is validated against a targeted subset of human RNA-seq data on chromosome 22. It successfully identifies the 54 high-confidence variants required to match the output of the original discovery engine.
```
wm75: Be more precise here and state the origin of that test data.

workflows/variant-calling/rna-variant-discovery/RNAvar.ga
```diff
@@ -0,0 +1,904 @@
...
+                "tool_shed": "toolshed.g2.bx.psu.edu"
+            },
+            "tool_state": "{\"__input_ext\": \"input\", \"chromInfo\": \"/opt/galaxy/tool-data/shared/ucsc/chrom/?.len\", \"input1\": {\"__class__\": \"ConnectedValue\"}, \"minhash\": false, \"prim_key_cond\": {\"prim_key_select\": \"\", \"__current_case__\": 0}, \"__page__\": 0, \"__rerun_remap_job_id__\": null}",
+            "tool_uuid": null,
+            "tool_version": "2.0.8",
+            "type": "tool",
+            "uuid": "27f1d0d4-b06c-4e39-a76b-a705964f4a90",
+            "when": null,
+            "workflow_outputs": []
+        },
```
wm75: The RNAStar Galaxy wrapper should already generate coordinate-sorted BAM output, so is this step necessary?

Please check and/or clarify.

workflows/variant-calling/rna-variant-discovery/RNAvar.ga
```diff
@@ -0,0 +1,904 @@
...
+                "tool_shed": "toolshed.g2.bx.psu.edu"
+            },
+            "tool_state": "{\"CN\": null, \"DS\": null, \"DT\": null, \"PI\": null, \"PL\": \"ILLUMINA\", \"PU\": \"run\", \"__input_ext\": \"input\", \"chromInfo\": \"/opt/galaxy/tool-data/shared/ucsc/chrom/?.len\", \"inputFile\": {\"__class__\": \"ConnectedValue\"}, \"read_group_id_conditional\": {\"do_auto_name\": true, \"__current_case__\": 0}, \"read_group_lb_conditional\": {\"do_auto_name\": true, \"__current_case__\": 0}, \"read_group_sm_conditional\": {\"do_auto_name\": true, \"__current_case__\": 0}, \"validation_stringency\": \"LENIENT\", \"__page__\": 0, \"__rerun_remap_job_id__\": null}",
+            "tool_uuid": null,
+            "tool_version": "3.1.1.0",
+            "type": "tool",
+            "uuid": "cf07f182-d8ab-4c43-8df3-9c263101cbe3",
+            "when": null,
+            "workflow_outputs": []
+        },
```
wm75: Try to understand if and how this step changes the data. Is there a way to use it in a meaningful way either now without having implemented sample sheet funtionality or maybe in the future?

workflows/variant-calling/rna-variant-discovery/RNAvar.ga
```diff
@@ -0,0 +1,904 @@
...
...se__\": 28, \"output\": [{\"__index__\": 0, \"type\": {\"type\": \"log\", \"__current_case__\": 0, \"input\": {\"__class__\": \"ConnectedValue\"}}}]}}, {\"__index__\": 1, \"software_cond\": {\"software\": \"fastqc\", \"__current_case__\": 8, \"output\": [{\"__index__\": 0, \"type\": \"data\", \"input\": {\"__class__\": \"ConnectedValue\"}}]}}, {\"__index__\": 2, \"software_cond\": {\"software\": \"picard\", \"__current_case__\": 17, \"output\": [{\"__index__\": 0, \"type\": \"markdups\", \"input\": {\"__class__\": \"ConnectedValue\"}}]}}, {\"__index__\": 3, \"software_cond\": {\"software\": \"snpeff\", \"__current_case__\": 26, \"input\": {\"__class__\": \"ConnectedValue\"}}}, {\"__index__\": 4, \"software_cond\": {\"software\": \"samtools\", \"__current_case__\": 24, \"output\": [{\"__index__\": 0, \"type\": {\"type\": \"flagstat\", \"__current_case__\": 1, \"input\": {\"__class__\": \"ConnectedValue\"}}}]}}, {\"__index__\": 5, \"software_cond\": {\"software\": \"gatk\", \"__current_case__\": 11, \"output\": [{\"__index__\": 0, \"type\": \"base_recalibrator\", \"input\": {\"__class__\": \"ConnectedValue\"}}]}}], \"title\": \"\", \"__page__\": 0, \"__rerun_remap_job_id__\": null}",
```
wm75: Please order the different MultiQC input reports in a logical way proceeding from raw read metrics through alignment, variant calling and annotation.

workflows/variant-calling/rna-variant-discovery/RNAvar.ga
```diff
@@ -0,0 +1,904 @@
...
...\"ch\"], \"HI_offset\": \"1\", \"outSAMprimaryFlag\": \"OneBestScore\", \"outSAMmapqUnique\": \"60\", \"wasp_conditional\": {\"waspOutputMode\": \"\", \"__current_case__\": 1}}, \"outWig\": {\"outWigType\": \"None\", \"__current_case__\": 0, \"outWigStrand\": \"false\"}, \"perf\": {\"outBAMsortingBinsN\": \"50\", \"winAnchorMultimapNmax\": \"50\"}, \"refGenomeSource\": {\"geneSource\": \"history\", \"__current_case__\": 1, \"genomeFastaFiles\": {\"__class__\": \"ConnectedValue\"}, \"genomeSAindexNbases\": \"14\", \"GTFconditional\": {\"GTFselect\": \"with-gtf\", \"__current_case__\": 0, \"sjdbGTFfile\": {\"__class__\": \"ConnectedValue\"}, \"sjdbGTFfeatureExon\": \"exon\", \"sjdbOverhang\": \"14\", \"quantmode_output\": {\"quantMode\": \"-\", \"__current_case__\": 0}}, \"diploidconditional\": {\"diploid\": \"No\", \"__current_case__\": 1}}, \"singlePaired\": {\"sPaired\": \"paired\", \"__current_case__\": 1, \"input1\": {\"__class__\": \"ConnectedValue\"}, \"input2\": {\"__class__\": \"ConnectedValue\"}}, \"twopass\": {\"twopassMode\": \"Basic\", \"__current_case__\": 1, \"twopass_read_subset\": \"\", \"sj_precalculated\": \"\"}, \"__page__\": 0, \"__rerun_remap_job_id__\": null}",
```
wm75: The primary advantage of building the STAR index on te fly (which otherwise is just very wasteful) is that you can control the `--sjdbOverhang` parameter (see https://nf-co.re/rnavar/1.2.3/docs/usage/#alignment-options).
This means that you need to expose the `sjdbOverhang` wrapper parameter as a workflow parameter with a helpful annotation.

lldelisle: For me `sjdbOverhang` is always exposed if you use a prebuilt genome reference without a prebuit gene model.

wm75: That's also a good reason, yes.
Do you think we should go this way @lldelisle ? Let the user choose from cached, prebuilt references, but bring their own gene model? Could be a very good compromise.

lldelisle: This is always the way I do, this is also the way it is done in the RNA-seq IWC workflow.
I think it should be cached prebuilt reference + own gtf.
Therefore I would add a step toolshed.g2.bx.psu.edu/repos/iuc/snpeff/snpEff_build_gb/5.4+galaxy0 to build the snpEff database matching the same reference fasta and the same gtf.
Using the cache option (reuse jobs with identical parameters) will allow to not really rebuild the database if it was already built with the same gtf.

workflows/variant-calling/rna-variant-discovery/README.md
```diff
@@ -0,0 +1,26 @@
+# RNA-seq Variant Discovery (RNAvar)
+
+This workflow is an implementation of a standard RNA-seq variant discovery pipeline. It handles the transition from raw reads to annotated variants using industry-standard tools.
+
+## Workflow Logic
+1. **Alignment:** **STAR** in 2-pass mode for high-accuracy splice-aware mapping.
+2. **Preprocessing:** **MarkDuplicates** and **GATK4 SplitNCigarReads** to prepare the BAM for variant calling by splitting reads into exon segments.
+3. **Recalibration:** **GATK4 BaseRecalibrator** (BQSR) using known polymorphic sites to adjust base quality scores.
+4. **Variant Calling:** **GATK4 HaplotypeCaller** with parameters specifically tuned for RNA-seq (e.g., ignoring soft-clipped bases).
+5. **Filtering:** **bcftools filter** applying hard filters (`FS > 30.0`, `QD < 2.0`) to reduce false positives.
```
wm75: These filters are user-configurable in the nf-core pipeline so should become workflow parameters, too.

workflows/variant-calling/rna-variant-discovery/RNAvar.ga
wm75: > This is always the way I do, this is also the way it is done in the RNA-seq IWC workflow.
> I think it should be cached prebuilt reference + own gtf.

Cool, sounds excellent to me, thanks!

In that case, however, how can we make sure that the reference fasta matches the one used for building the STAR index?

Also,

> Therefore I would add a step toolshed.g2.bx.psu.edu/repos/iuc/snpeff/snpEff_build_gb/5.4+galaxy0 to build the snpEff database matching the same reference fasta and the same gtf.

I'm not so convinced this is a good idea, building SnpEff annotation files from gff is a fragile art for eukaryotic genomes and the pre-built cached ones might contain fewer errors.

So, essentially, the question to answer is: what's the least error-prone way to have matching genomes at the STAR, GATK and SnpEff steps.

Potential differences between SnpEff and custom gff annotations are another thing, but wouldn't be catastrophic for the workflow. Users who have a carefully crafted SnpEff db based specifically for their gff could also ask for server-side installation?
For exotic cases where there's no reasonably close existing SnpEff db, we *could* offer an optional branch of the WF to build the db from the reference and the gff, I just wouldn't make this the default.

Of course, I'm only making all of this up while typing so feel free to disagree with me :-)

workflows/variant-calling/rna-variant-discovery/RNAvar.ga
emregulben: This step is required for the GATK4 tools used later in the workflow. Tools like HaplotypeCaller and BaseRecalibrator are designed to look for 'Read Group' metadata in the BAM header. Without these tags, the tools throw an error and won't process the file (as noted in the [GATK documentation on Read Groups](https://gatk.broadinstitute.org/hc/en-us/articles/360035890671-Read-groups)).

By including this step now with default values, we ensure the workflow is robust and GATK-compliant. It also sets us up nicely for the future. If we implement sample sheet support later, this tool is exactly where that metadata would be injected into the pipeline.

workflows/variant-calling/rna-variant-discovery/README.md
emregulben: Done: https://github.com/galaxyproject/iwc/pull/1188/commits/5146320e898fd1b092f01e231b5015a2a9a732f8 ✅

workflows/variant-calling/rna-variant-discovery/RNAvar.ga
emregulben: https://github.com/galaxyproject/iwc/pull/1188/commits/5146320e898fd1b092f01e231b5015a2a9a732f8 ✔️

workflows/variant-calling/rna-variant-discovery/RNAvar.ga
emregulben: I verified STAR already handles the sorting. Removed the step: https://github.com/galaxyproject/iwc/pull/1188/commits/5146320e898fd1b092f01e231b5015a2a9a732f8

workflows/variant-calling/rna-variant-discovery/README.md
emregulben: Added the limitations section: https://github.com/galaxyproject/iwc/pull/1188/commits/5146320e898fd1b092f01e231b5015a2a9a732f8

workflows/variant-calling/rna-variant-discovery/README.md
emregulben: Done here: https://github.com/galaxyproject/iwc/pull/1188/commits/5146320e898fd1b092f01e231b5015a2a9a732f8

workflows/variant-calling/rna-variant-discovery/RNAvar.ga
emregulben: I’ve exposed `sjdbOverhang` as a mandatory workflow parameter. Regarding the SnpEff database discussion, I’ve kept the database selection as a string input to use pre-built/cached local versions, avoiding the potential fragility of building custom databases from GFFs during the run. Let me know if the current version is fine. https://github.com/galaxyproject/iwc/pull/1188/commits/5146320e898fd1b092f01e231b5015a2a9a732f8

workflows/variant-calling/rna-variant-discovery/RNAvar.ga
```diff
@@ -0,0 +1,972 @@
...
+            ],
+            "label": "SnpEff Genome Version",
+            "name": "Input parameter",
+            "outputs": [],
+            "position": {
+                "left": 0,
+                "top": 209.96850976711244
+            },
+            "tool_id": null,
+            "tool_state": "{\"multiple\": false, \"validators\": [], \"parameter_type\": \"text\", \"optional\": false}",
```
lldelisle: You need to force to choose among the prebuit SnpEff databases:

Image

With attempt restriction based on connections

lldelisle:
Here is my proposed variation: https://usegalaxy.eu/u/delislel/w/rna-seq-variant-discovery-ld-version
What I changed:
- use as input fastqs: PE collection, I think it is easier if you want to process multiple samples at the time and even if you don't this ensure that all your output files will have the name of your sample.
- use built-in genome instead of fasta
- add an output to STAR (genecounts) for me this is very useful to see this statistics in the MultiQC to check you put the right GTF (you should have a high proportion of reads mapping to  genes if the GTF is correct).
- I also rearranged the inputs to they are in the order they are used in the pipeline
- Finally, I put 'hidden' all intermediate steps

Tell me what you think and feel free to change it.

wm75:
> Here is my proposed variation: https://usegalaxy.eu/u/delislel/w/rna-seq-variant-discovery-ld-version

Wow, thanks @lldelisle and sorry for not getting round to this again earlier.

@emregulben I like Lucille's version a lot and she gives a detailed list of what she's been changing compared to your version.

Now I've built further (https://usegalaxy.eu/published/workflow?id=c1b8da2f8da2a30a) on her flavor of the WF and:
- added the equivalent of the nf-core pipeline's exon_bed input (called it "Variant discovery regions"). 

  Currently, this gets handled just a little different from what nf-core does. When you don't provide it, genomic variant calling gets performed, while in that case the nf-core version constructs the file from your gff and has an extra flag for doing genomic calling. I leave it up to you whether you want to try and implement that exact behavior or whether you accept that bit of difference and document it in the README.

If you continue working from this version (my suggestion), then you mostly only need to put the (adjusted) WF readme and the release number back in, and adjust the tests to work with collections (there are good examples in this repo). Also, please add Lucille to the WF authors like [here](https://github.com/galaxyproject/iwc/blob/7ff26e9b4452a81c309a8607b5980f79e042d968/workflows/transcriptomics/brew3r/BREW3R.ga#L7-L9) and to the dockstore.yml, too.

workflows/variant-calling/rna-variant-discovery/README.md
```diff
@@ -0,0 +1,37 @@
...
+6. **Annotation:** Functional annotation of variants using **SnpEff** against a selected local database.
+
+## Inputs
+* **Forward Reads (R1)**: Fastq sequencing reads (forward). Supports gzipped (`.gz`) files.
+* **Reverse Reads (R2)**: Fastq sequencing reads (reverse). Supports gzipped (`.gz`) files.
```
wm75: This will need to change when using a list of pairs input.

workflows/variant-calling/rna-variant-discovery/README.md
```diff
@@ -0,0 +1,37 @@
...
+* **Reference Genome**: FASTA file of the reference genome (e.g., hg38).
```
wm75: Would become a cached genome on the server

workflows/variant-calling/rna-variant-discovery/README.md
```diff
@@ -0,0 +1,37 @@
...
+* **Genome Annotation**: GFF3 file containing gene models for splice-aware alignment.
+* **Known Variants (dbSNP)**: VCF file of known polymorphisms for base quality score recalibration.
+* **Known Indels**: VCF file of known insertions and deletions for base quality score recalibration.
+* **SnpEff Genome Version**: The name of the locally installed SnpEff database (e.g., hg38).
+* **Variant Exclusion Filter**: A bcftools-style expression for hard-filtering variants (default: FS > 30.0 || QD < 2.0).
+* **STAR Genome Overhang**: Sequence length around junctions for STAR indexing. Set to (Read Length - 1). For example, use 99 for 100bp reads.
```
wm75, line 17: Reorder inputs according to latest WF version

workflows/variant-calling/rna-variant-discovery/README.md
```diff
@@ -0,0 +1,37 @@
...
+5. **Filtering:** **bcftools filter** applying hard filters (`FS > 30.0`, `QD < 2.0`) to reduce false positives.
+6. **Annotation:** Functional annotation of variants using **SnpEff** against a selected local database.
+
+## Inputs
+* **Forward Reads (R1)**: Fastq sequencing reads (forward). Supports gzipped (`.gz`) files.
+* **Reverse Reads (R2)**: Fastq sequencing reads (reverse). Supports gzipped (`.gz`) files.
+* **Reference Genome**: FASTA file of the reference genome (e.g., hg38).
+* **Genome Annotation**: GFF3 file containing gene models for splice-aware alignment.
+* **Known Variants (dbSNP)**: VCF file of known polymorphisms for base quality score recalibration.
+* **Known Indels**: VCF file of known insertions and deletions for base quality score recalibration.
```
wm75: document here also the new exon_bed-like input

workflows/variant-calling/rna-variant-discovery/README.md
```diff
@@ -0,0 +1,37 @@
+# RNA-seq Variant Discovery (RNAvar)
+
+This workflow is a faithful reimplementation of the [nf-core/rnavar](https://nf-co.re/rnavar) pipeline for the Galaxy ecosystem. It handles the transition from raw reads to annotated variants using industry-standard tools and follows the established best practices for RNA-seq variant discovery.
```
wm75:
```suggestion
This workflow is an attempt at a faithful reimplementation of the [nf-core/rnavar](https://nf-co.re/rnavar) pipeline for the Galaxy ecosystem. nf-core rnavar handles the transition from raw reads to annotated variants using industry-standard tools and follows established best practices for RNA-seq variant discovery. This implementation for Galaxy still has a few limitations (see below).
```

workflows/variant-calling/rna-variant-discovery/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Workflow Logic
+1. **Alignment:** **STAR** in 2-pass mode for high-accuracy splice-aware mapping.
+2. **Preprocessing:** **MarkDuplicates** and **GATK4 SplitNCigarReads** to prepare the BAM for variant calling by splitting reads into exon segments.
+3. **Recalibration:** **GATK4 BaseRecalibrator** (BQSR) using known polymorphic sites to adjust base quality scores.
+4. **Variant Calling:** **GATK4 HaplotypeCaller** with parameters specifically tuned for RNA-seq (e.g., ignoring soft-clipped bases).
+5. **Filtering:** **bcftools filter** applying hard filters (`FS > 30.0`, `QD < 2.0`) to reduce false positives.
```
wm75:
```suggestion
5. **Filtering:** called variants are soft-filtered with **bcftools filter** (default filter expression: `FS > 30.0`, `QD < 2.0`) to reduce false positives.
```

workflows/variant-calling/rna-variant-discovery/README.md
```diff
@@ -0,0 +1,37 @@
...
+* **MultiQC Quality Report**: An HTML report aggregating quality metrics from alignment and preprocessing steps.
+
+## Testing and Parity
+The workflow is validated against a targeted subset of human RNA-seq data on chromosome 22, sourced from the [nf-core/rnavar test dataset](https://github.com/nf-core/test-datasets/tree/rnavar). It achieves technical parity with the reference discovery engine, identifying 54 high-confidence variants.
+
+## Limitations
+Compared to the full [nf-core/rnavar (v1.2.3)](https://nf-co.re/rnavar/1.2.3/docs/usage/) implementation, this Galaxy workflow currently has the following limitations:
+* **Sample Handling:** Does not support CSV sample sheets; inputs must be provided as individual datasets or collections.
+* **Modularity:** Base recalibration (BQSR) and variant filtering are mandatory steps, whereas they are optional in the original pipeline.
+* **Preprocessing:** No current support for UMI-Tools-based deduplication or optional read trimming.
```
wm75, line 36: Also worth mentioning:

- the Variant calling step doesn't let you configure --gatk_hc_call_conf
- Filtering uses bcftools filter and doesn't offer clustered SNPs filtering

emregulben:
@wm75 I've updated the test file so the input data correctly uses collections. I also added the `regions.bed` file to ensure the workflow's optional filtering step is properly tested. Additionally, I had to update a line in the `mills_and_1000G.indels.vcf` file to adjust the chromosome length. Since the workflow uses the hg38 reference genome, the tools require the actual biological chromosome sizes instead of the artificially tiny ones from the old test data. Because the pipeline is now validating against this full reference, GATK automatically includes the complete list of human chromosomes in the header of the final output file. Because of this, I had to change the expected line count for the test assertion.

Here's the final workflow: https://usegalaxy.eu/u/emreg/w/rna-seq-variant-discovery

wm75:
The test now fails because we cannot allocate enough memory for a STAR job mapping against the human genome.

I guess the correct way forward would be to implement https://nf-co.re/rnavar/1.2.3/parameters/#skip_baserecalibration

Then you could just reuse all the yeast datasets from @lldelisle 's https://github.com/galaxyproject/iwc/blob/main/workflows/transcriptomics/rnaseq-pe/rnaseq-pe-tests.yml and sitwch the test to mapping against the *S. cerevisiae* genome.

wm75:
see also the note at: https://nf-co.re/rnavar/1.2.3/docs/usage/#postprocessing-of-alignment-output for why you'd want to have that option implemented for wider organism support anyway and for the tests specifically.

emregulben:
Updated the tests to use yeast data: https://github.com/galaxyproject/iwc/pull/1188/commits/20cf3b2a5c825810dba369c465930af67d8bfb96

PR #1206, opened 2026-04-09
Add histological staining area quantification workflow

dianichj:
This workflow quantifies the area of histological staining in brightfield 
histological images using colour deconvolution and automated thresholding. 
Compatible with Masson's Trichrome (MT) and Immunohistochemistry (IHC).

Published workflow: https://usegalaxy.eu/u/dianitachj24/w/histological-staining-area-quantification

### Status
Draft PR - test data and .dockstore.yml will be added before final merge.

- [x] test-data to Zenodo (files > 100 KB)
- [x] CHANGELOG.md
- [x] .dockstore.yml

cc: @mvdbeek @afgane


dianichj:
Hi @mvdbeek, the workflow test is failing due to an error in the `imagej2_analyze_particles_binary` tool (changeset `862af85a50ec`):

```
closing tag '[/tmp/shed_dir/.../imagej2_analyze_particles_binary_jython_script.py, ...]' at position 53 doesn't match any open tag
```

Could it be a bug in the tool rather than in the workflow or test data? The step (`Generate ROIs of stained region`) is an auxiliary step for visual validation step, for example if one would like to upload data to OMERO or automatically generate all the ROIs. 

Would you recommend removing this step from the workflow for now, or is this a known issue that can be worked around? Other than that, I think the PR looks okay. Happy and grateful to follow your guidance before merging :)!

mvdbeek:
This is a planemo bug, working on it. However this is triggered by a failed job, so once the fix is in the cause should be clear(er).

mvdbeek:
With the planemo bug fixed ([planemo PR for the rich-markup escaping](https://github.com/galaxyproject/planemo/), to be filed), the real failure of Step 14 (`Generate ROIs of stained region`, `imgteam/imagej2_analyze_particles_binary`@`862af85a50ec`) is now visible in the latest CI run. All three jobs fail identically:

```
Traceback (most recent call last):
  File ".../imagej2_analyze_particles_binary_jython_script.py", line 66, in <module>
    for i, roi in enumerate(ov):
TypeError: 'NoneType' object is not iterable
```

### Root cause

In `imagej2_analyze_particles_binary_jython_script.py` (toolshed rev `862af85a50ec`), when `export_rois=true` (which this workflow sets), the script runs `Analyze Particles…` with `show=Overlay` and then does:

```python
ov = input_image_plus_copy.getOverlay()
with open(roi_coordinate_file, 'w') as fo:
    fo.write("shape\tpoints\tlabel\tt\tz\n")
    for i, roi in enumerate(ov):   # crashes when ov is None
        ...
```

ImageJ's `getOverlay()` returns `None` when zero particles match the criteria (here `size=50-Infinity`, `circularity=0.0–1.0`, `black_background=yes`). So for each of the three test inputs the upstream binary mask contains no blob ≥50 px² that the analyzer accepts, and the script crashes instead of producing a header-only ROI table.

The `ImageJ.cfg: Permission denied` line in stderr is unrelated noise — ImageJ continues past it.

### Suggested next steps

1. **Tool-side fix** (upstream `imgteam/imagej2_analyze_particles_binary`): guard the iteration, e.g.
   ```python
   ov = input_image_plus_copy.getOverlay()
   with open(roi_coordinate_file, 'w') as fo:
       fo.write("shape\tpoints\tlabel\tt\tz\n")
       if ov is not None:
           for i, roi in enumerate(ov):
               ...
   ```
   Two-line fix plus a regression test with an input that has no particles ≥ threshold.
2. **Workflow-side check**: verify the thresholded masks for the three Zenodo TIFFs do actually contain stained regions ≥50 px². If they do, the bug is solely in the wrapper; if they don't, either the size threshold or the upstream thresholding step needs adjusting for the test data.

dianichj:
> With the planemo bug fixed ([planemo PR for the rich-markup escaping](https://github.com/galaxyproject/planemo/), to be filed), the real failure of Step 14 (`Generate ROIs of stained region`, `imgteam/imagej2_analyze_particles_binary`@`862af85a50ec`) is now visible in the latest CI run. All three jobs fail identically:
> 
> ```
> Traceback (most recent call last):
>   File ".../imagej2_analyze_particles_binary_jython_script.py", line 66, in <module>
>     for i, roi in enumerate(ov):
> TypeError: 'NoneType' object is not iterable
> ```
> 
> ### Root cause
> In `imagej2_analyze_particles_binary_jython_script.py` (toolshed rev `862af85a50ec`), when `export_rois=true` (which this workflow sets), the script runs `Analyze Particles…` with `show=Overlay` and then does:
> 
> ```python
> ov = input_image_plus_copy.getOverlay()
> with open(roi_coordinate_file, 'w') as fo:
>     fo.write("shape\tpoints\tlabel\tt\tz\n")
>     for i, roi in enumerate(ov):   # crashes when ov is None
>         ...
> ```
> 
> ImageJ's `getOverlay()` returns `None` when zero particles match the criteria (here `size=50-Infinity`, `circularity=0.0–1.0`, `black_background=yes`). So for each of the three test inputs the upstream binary mask contains no blob ≥50 px² that the analyzer accepts, and the script crashes instead of producing a header-only ROI table.
> 
> The `ImageJ.cfg: Permission denied` line in stderr is unrelated noise — ImageJ continues past it.
> 
> ### Suggested next steps
> 1. **Tool-side fix** (upstream `imgteam/imagej2_analyze_particles_binary`): guard the iteration, e.g.
>    ```python
>    ov = input_image_plus_copy.getOverlay()
>    with open(roi_coordinate_file, 'w') as fo:
>        fo.write("shape\tpoints\tlabel\tt\tz\n")
>        if ov is not None:
>            for i, roi in enumerate(ov):
>                ...
>    ```
>    
>    
>        
>          
>        
>    
>          
>        
>    
>        
>      
>    Two-line fix plus a regression test with an input that has no particles ≥ threshold.
> 2. **Workflow-side check**: verify the thresholded masks for the three Zenodo TIFFs do actually contain stained regions ≥50 px². If they do, the bug is solely in the wrapper; if they don't, either the size threshold or the upstream thresholding step needs adjusting for the test data.

Thanks a lot @mvdbeek for the detailed diagnosis! The `size=50-Infinity` threshold is indeed too strict for these test images. The data is CD11b IHC from mouse cardiac cryosections after MI and CD11b marks myeloid cells, which are sparse in the heart even after infarction, so the stained regions per ROI are small. I am re-running the workflow in Galaxy with a lower size threshold to confirm that the stained regions are correctly captured by the Analyze Particles step, and will update the `.ga` accordingly. That should fix the error, otherwise I would select different test images, there might be the current test ones are not representative enough.

dianichj:
Hi @mvdbeek, I think the PR is ready. The workflow test passes (`All 1 test(s) executed passed`), and the only remaining failure is the planemo report generation error you already flagged:

```
TypeError: 'NoneType' object is not iterable
```

I also adjusted the `size` threshold in the Analyze Particles step from `50-Infinity` to `2-Infinity` to better reflect the sparse CD11b myeloid cell staining in cardiac tissue, and updated the test images to IZ ROIs which have more detectable staining. Please let me know if anything else is needed and thanks a lot!

mvdbeek (approved):
Let's get this in, the particle analyzer tool needs an update though to not crash when no overlay is found.

PR #1217, opened 2026-04-21
Convert duplicate-keyed asserts blocks to list form

jmchilton:
## Summary

YAML silently collapses duplicate mapping keys, so `asserts:` blocks that repeat
an assertion type (`has_text:`, `has_line:`, `has_text_matching:`) were only
running the **last** assertion per type — every earlier assertion was silently
dropped by the YAML loader and never executed.

This converts 12 affected workflow-tests files from duplicate-keyed dict form
to list form so every assertion actually runs:

```yaml
# Before — second has_text: overwrites first, "gene" assertion never ran
asserts:
  has_text:
    text: "gene"
  has_text:
    text: "BUSTED"

# After — both assertions run
asserts:
  - that: has_text
    text: "gene"
  - that: has_text
    text: "BUSTED"
```

## ⚠ Heads-up for maintainers: semantic change

Because the first N−1 assertions per type were previously dead code, some of
these tests may have been silently passing on assertion text that doesn't
actually match current tool output. Running them for real may surface
legitimate assertion failures. That's not a regression in this PR — it's the
first time those assertions have run. If CI fails on one of these, the fix is
to update the expected text (the author's intent was clearly to assert both),
not to revert this PR.

## Files changed (12)

- `VGP-assembly-v2/hi-c-contact-map-for-assembly-manual-curation/…-tests.yml`
- `VGP-assembly-v2/kmer-profiling-hifi-trio-VGP2/…-tests.yml`
- `bacterial_genomics/bacterial-quality-and-contamination-control-post-assembly/…-tests.yml`
- `comparative_genomics/hyphy/capheine-core-and-compare-tests.yml`
- `genome-assembly/bacterial-genome-assembly/…-tests.yml`
- `genome-assembly/quality-and-contamination-control-raw-reads/…-tests.yml`
- `microbiome/host-contamination-removal/host-contamination-removal-long-reads/…-tests.yml`
- `microbiome/host-contamination-removal/host-contamination-removal-short-reads/…-tests.yml`
- `read-preprocessing/short-read-qc-trimming/…-tests.yml`
- `sars-cov-2-variant-calling/sars-cov-2-pe-illumina-artic-variant-calling/…-tests.yml`
- `sars-cov-2-variant-calling/sars-cov-2-se-illumina-wgs-variant-calling/…-tests.yml`
- `variant-calling/ploidy-aware-genotype-calling/…-test.yml`

## How this was found

Galaxy's new `gxwf validate-tests` workflow-test validator (in-progress work on
the Galaxy side) swept all IWC workflow-tests files against the Pydantic
`Tests` model. `ordered_load` in Galaxy's YAML helper raises on duplicate
mapping keys (per YAML 1.2), which surfaced these authoring bugs.

A follow-up PR will handle the remaining cases (duplicate `text:` / `expression:`
*inside* a single assertion item, plus a few misc structural dupes like
duplicate output names and `assert:` vs `asserts:` typos).

FOR CONTRIBUTOR:

FOR REVIEWERS:

### DIFF
--- a/workflows/VGP-assembly-v2/hi-c-contact-map-for-assembly-manual-curation/hi-c-map-for-assembly-manual-curation-tests.yml
+++ b/workflows/VGP-assembly-v2/hi-c-contact-map-for-assembly-manual-curation/hi-c-map-for-assembly-manual-curation-tests.yml
@@ -89,9 +89,9 @@
           delta: 5000
     P telomeres bed:
       asserts:
-        has_text:
+        - that: has_text
           text: "scaffold_10.H2"
-        has_text:
+        - that: has_text
           text: "scaffold_1.H1"
     Pretext All tracks:
       asserts:
@@ -224,9 +224,9 @@
           delta: 10000
     P telomeres bed:
       asserts:
-        has_text:
+        - that: has_text
           text: "scaffold_10.H2"
-        has_text:
+        - that: has_text
           text: "scaffold_1.H1"
     Merged Hi-C Alignments on Scaffolds:
       asserts:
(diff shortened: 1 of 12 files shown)


jmchilton:
I'm going to close this out and rebase into a green PR that can be merged and red tests we can track down.

PR #1252, opened 2026-06-03
fastq-to-matrix-10x-v3: hardcode soloBarcodeReadLength=0 for v3 chemistry

scottcain:
## Problem

The `fastq-to-matrix-10x-v3` workflow exposes `soloBarcodeReadLength` as a user-facing boolean parameter (step 4, "Barcode Size is same size of the Read"). For all standard 10x v3 libraries the barcode does not occupy the full read length, so this must always be `0` (disabled).

In practice the `ConnectedValue` routing does not reliably deliver the user's selection to STARsolo. During testing, both `true` and `false` UI settings produced identical, incorrect output: STARsolo log showed `Solo: CB 17bp` instead of `Solo: CB 16bp`, meaning barcode detection was wrong regardless of what the user selected. Direct STARsolo tool runs with `soloBarcodeReadLength=0` produced correct output.

The test YAML explicitly passes `Barcode Size is same size of the Read: false`, which is why CI passes, but interactive users reliably hit the bug.

## Fix

- Remove the `parameter_input` step entirely (step 4)
- Remove the `sc|soloBarcodeReadLength` connection from the STARsolo step
- Hardcode `"soloBarcodeReadLength": "0"` directly in the STARsolo `tool_state`
- Remove the now-nonexistent parameter from the test YAML

This approach is unambiguous: `soloBarcodeReadLength` is always `0` for v3 chemistry and there is no reason to expose it as a user-adjustable parameter in a workflow named for v3.

## Evidence

Reanalysis of Wang et al. 2024 (*Aedes aegypti* midgut scRNA-seq, GEO GSE246612) on usegalaxy.org. Workflow invocations `a9a0b6a0f307171b` (step 4 = true) and `983ee63c0b6a234a` (step 4 = false) both failed identically with `Solo: CB 17bp`.

## Testing

Existing test YAML updated to remove the deleted parameter. Test assertions are unchanged; expected outputs are unchanged since the correct value (`0`) was already being passed in the test.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/workflows/scRNAseq/fastq-to-matrix-10x/scrna-seq-fastq-to-matrix-10x-v3-tests.yml
+++ b/workflows/scRNAseq/fastq-to-matrix-10x/scrna-seq-fastq-to-matrix-10x-v3-tests.yml
@@ -23,7 +23,6 @@
           - hash_function: SHA-1
             hash_value: bba127e369535d226290a74858efafd7f359168e
     reference genome: dm6
-    Barcode Size is same size of the Read: false
     gtf:
       class: File
       location: https://zenodo.org/record/6457007/files/Drosophila_melanogaster.BDGP6.32.109_UCSC.gtf.gz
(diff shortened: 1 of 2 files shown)


scottcain:
It's not clear to me that this is the right fix; I'll re-evaluate and resubmit after further consideration.

lldelisle:
Thanks for noticing, I will check, maybe one of the automatic update broke it.

lldelisle:
(I put it as parameter as some sequencing facilities only provide the length of the barcode, therefore it is a good way to check you gave the good R1/R2).

PR #1063, opened 2026-01-15
Add workflow for metagenomic taxonomic community profiling

bebatut:
ping @minamehr

FOR CONTRIBUTOR:
* [X] I have read the [Adding workflows guidelines](https://github.com/galaxyproject/iwc/blob/main/workflows/README.md#adding-workflows)
* [X] License permits unrestricted use (educational + commercial)
* [ ] Please also take note of the reviewer guidelines below to facilitate a smooth review process.

FOR REVIEWERS:
* [ ] .dockstore.yml: file is present and aligned with creator metadata in workflow. ORCID identifiers are strongly encouraged in creator metadata. The .dockstore.yml file is required to run tests
* [ ] Workflow is sufficiently generic to be used with lab data and does not hardcode sample names, reference data and can be run without reading an accompanying tutorial.
* [ ] In workflow: annotation field contains short description of what the workflow does. Should start with `This workflow does/runs/performs … xyz … to generate/analyze/etc …`
* [ ] In workflow: workflow inputs and outputs have human readable names (spaces are fine, no underscore, dash only where spelling dictates it), no abbreviation unless it is generally understood. Altering input or output labels requires adjusting these labels in the the workflow-tests.yml file as well
* [ ] In workflow: `name` field should be human readable (spaces are fine, no underscore, dash only where spelling dictates it), no abbreviation unless generally understood
* [ ] Workflow folder: prefer dash (`-`) over underscore (`_`), prefer all lowercase. Folder becomes repository in [iwc-workflows organization](https://github.com/iwc-workflows) and is included in TRS id
* [ ] Readme explains what workflow does, what are valid inputs and what outputs users can expect. If a tutorial or other resources exist they can be linked. If a similar workflow exists in IWC readme should explain differences with existing workflow and when one might prefer one workflow over another
* [ ] Changelog contains appropriate entries
* [ ] Large files (> 100 KB) are uploaded to zenodo and location urls are used in test file

PR #1257, opened 2026-06-19
Add SameStr workflow

xens25:
FOR CONTRIBUTOR:

FOR REVIEWERS:


workflows/microbiome/samestr-workflow/.dockstore.yml
```diff
@@ -0,0 +1,10 @@
+version: 1.2
+workflows:
+- name: main
+  subclass: Galaxy
+  publish: true
+  primaryDescriptorPath: /samestr-workflow.ga
+  testParameterFiles:
+   - /samestr-workflow-tests.yml
+  authors:
+  - name: 'Xenia Morera Martinez'
```
mvdbeek: Can you please put your ORCID id here ?

workflows/microbiome/samestr-workflow/samestr-workflow.ga
```diff
@@ -0,0 +1,2230 @@
...
+        }
+    ],
+    "creator": [
+        {
+            "class": "Person",
+            "familyName": "Morera Mart\u00ednez",
+            "name": "Xenia "
+        },
+        {
+            "class": "Person",
```
mvdbeek: this doesn't match the authors in .dockstore.yml, but please also add the orcid ids here

workflows/microbiome/samestr-workflow/samestr-workflow.ga
```diff
@@ -0,0 +1,2230 @@
...
+                }
+            },
+            "tool_id": "toolshed.g2.bx.psu.edu/repos/iuc/kneaddata/kneaddata/0.12.1+galaxy1",
+            "tool_shed_repository": {
+                "changeset_revision": "10f8abc1fbdf",
+                "name": "kneaddata",
+                "owner": "iuc",
+                "tool_shed": "toolshed.g2.bx.psu.edu"
+            },
+            "tool_state": "{\"alignment\": {\"alignment_tool\": {\"tool_choice\": \"bowtie2\", \"__current_case__\": 0, \"bowtie2_db\": \"hg19\"}, \"save_contaminants\": false}, \"fastqc\": {\"fastqc_report\": {\"select_fastqc\": \"no_fastqc\", \"__current_case__\": 1}, \"run_trim_repetitive\": false}, \"log_level\": \"INFO\", \"read_type\": {\"select_read_type\": \"paired\", \"__current_case__\": 1, \"save_unmatched\": true, \"paired_collection\": {\"__class__\": \"ConnectedValue\"}}, \"tandem\": {\"trf_step\": {\"trf_bool\": \"skip\", \"__current_case__\": 1}}, \"trimmomatic\": {\"trimmomatic_step\": {\"decide_trimmomatic\": \"run_trimmomatic\", \"__current_case__\": 0, \"sequencer\": \"NexteraPE\"}}, \"__page__\": 0, \"__rerun_remap_job_id__\": null}",
```
mvdbeek: Can you please make the bowtie database configurable ?

workflows/microbiome/samestr-workflow/README.md
```diff
@@ -0,0 +1,19 @@
+# SameStr workflow for shared-strain detection in metagenomes
+
+This workflow processes paired-end shotgun metagenomic sequencing reads from multiple samples. It performs the following steps:
+
+- **Preprocessing**: Quality trimming and host read removal using KneadData
+- **Taxonomic profiling**: MetaPhlAn4 or mOTUs for taxonomic classification
```
mvdbeek: Is this different from Metaphlan below ?

workflows/microbiome/samestr-workflow/samestr-workflow.ga
```diff
@@ -0,0 +1,2230 @@
...
+                            "owner": "bgruening",
+                            "tool_shed": "toolshed.g2.bx.psu.edu"
+                        },
+                        "tool_state": "{\"algorithm_advanced_options\": {\"advanced_options\": {\"customize_algorithm_options\": \"default_options\", \"__current_case__\": 1}}, \"all_taxonomic_level\": false, \"biom_format\": false, \"cami_format_option\": {\"cami_options\": \"no\", \"__current_case__\": 1}, \"full_rank_taxonomy\": false, \"full_species_name\": false, \"inputs\": {\"input_type\": {\"selector\": \"reads\", \"__current_case__\": 0, \"reads_type\": {\"selector\": \"paired\", \"__current_case__\": 0, \"forward_reads\": {\"__class__\": \"ConnectedValue\"}, \"reverse_reads\": {\"__class__\": \"ConnectedValue\"}}}, \"sample_name\": null, \"db\": {\"db_source\": \"cached\", \"__current_case__\": 0, \"db_cached\": \"db_from_2026-04-27T094930Z\"}}, \"ncbi_taxonomy_ids\": false, \"ref_motus\": false, \"result_as_counts\": false, \"save_bam\": false, \"save_mgc\": false, \"taxonomic_level_option\": {\"taxonomic_option\": \"no\", \"__current_case__\": 1}, \"__page__\": 0, \"__rerun_remap_job_id__\": null}",
```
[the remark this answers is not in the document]
mvdbeek: This one too should be configurable

workflows/microbiome/samestr-workflow/samestr-workflow.ga
mvdbeek: It also seems like `sequencer` is something to make configurable ?

workflows/microbiome/samestr-workflow/README.md
xens25: No, it's the same step. I'll add them together

xens25:
@mvdbeek Thanks for the suggestions. I added the fixes, let me know if there's anything else that needs to be modified

mvdbeek:
I see you've made the parameters runtime values, but can you please make them workflow input parameters ? Here's a tutorial that hopefully covers how to do this: https://training.galaxyproject.org/training-material/topics/galaxy-interface/tutorials/workflow-parameters/tutorial.html

xens25:
@mvdbeek Thanks for the tutorial, it was very helpful. I'm not sure if it's now correct, I've added the parameters as text inputs in both the subworkflows and workflow.

xens25:
I'm not sure why the test is failing, in the terminal both lint and test pass. I think it might be related to the subworkflows. @mvdbeek could you provide some insight on how to fix this?

mvdbeek:
Most of the databases and database values that the workflow uses are only available at usegalaxy.eu and are not portable to other Galaxy instances. We load the distributed CVMFS file system in the tests so that anyone can use these workflows. I'll see that we can sync these but that will take a couple of days.

I would also strongly suggest using this for all databases instead of pre-specifying the valid options:
Screenshot 2026-07-03 at 09 25 50

xens25:
Thanks @mvdbeek ! I’ve updated the database option. Let me know if any further fixes are needed from my side.

xens25:
> Most of the databases and database values that the workflow uses are only available at usegalaxy.eu and are not portable to other Galaxy instances. We load the distributed CVMFS file system in the tests so that anyone can use these workflows. I'll see that we can sync these but that will take a couple of days.
> 
> I would also strongly suggest using this for all databases instead of pre-specifying the valid options: Screenshot 2026-07-03 at 09 25 50

Do we have any updates on this?

mvdbeek:
working on that here: https://github.com/galaxyproject/idc/pull/89.
most of the databases are there but still working on samestr

workflows/microbiome/samestr-workflow/README.md
```diff
@@ -0,0 +1,18 @@
+# SamestrGal for shared-strain detection in metagenomes
```
wm75: Not sure that's the best name for the workflow. Other WFs do not mention Galaxy in their name because it's rather obvious after all.

workflows/microbiome/samestr-workflow/samestrgal-tests.yml
```diff
@@ -0,0 +1,90 @@
+- doc: Test outline for samestrgal
+  job:
+    Raw Reads:
+      class: Collection
+      collection_type: list:paired
+      elements:
+      - class: Collection
+        type: paired
```
wm75:
```suggestion
        collection_type: paired
```

see the recent #1292

workflows/microbiome/samestr-workflow/samestrgal-tests.yml
```diff
@@ -0,0 +1,90 @@
...
+        - class: File
+          identifier: forward
+          location: https://zenodo.org/records/20745835/files/28C_R1.fastq.gz
+          filetype: fastqsanger.gz
+        - class: File
+          identifier: reverse
+          location: https://zenodo.org/records/20745835/files/28C_R2.fastq.gz
+          filetype: fastqsanger.gz
+      - class: Collection
+        type: paired
```
wm75:
```suggestion
        collection_type: paired
```

workflows/microbiome/samestr-workflow/samestrgal-tests.yml
```diff
@@ -0,0 +1,90 @@
...
+        - class: File
+          identifier: forward
+          location: https://zenodo.org/records/20745835/files/28B_R1.fastq.gz
+          filetype: fastqsanger.gz
+        - class: File
+          identifier: reverse
+          location: https://zenodo.org/records/20745835/files/28B_R2.fastq.gz
+          filetype: fastqsanger.gz
+      - class: Collection
+        type: paired
```
wm75:
```suggestion
        collection_type: paired
```

workflows/microbiome/samestr-workflow/samestrgal-tests.yml
```diff
@@ -0,0 +1,90 @@
...
+        has_text:
+          text: "28C"
+    "Co-occurrences (MetaPhlAn)":
+      asserts:
+        has_text:
+          text: "28A"
+        has_text:
+          text: "28B"
+        has_text:
+          text: "28C"
```
wm75: please consult https://github.com/galaxyproject/iwc/pull/1275 and fix the asserts accordingly

workflows/microbiome/samestr-workflow/samestrgal.ga
```diff
@@ -0,0 +1,1299 @@
...
+                        "post_job_actions": {},
+                        "tool_id": "__RELABEL_FROM_FILE__",
+                        "tool_state": "{\"how\": {\"how_select\": \"txt\", \"__current_case__\": 0, \"labels\": {\"__class__\": \"ConnectedValue\"}, \"strict\": false}, \"input\": {\"__class__\": \"ConnectedValue\"}, \"__page__\": 0, \"__rerun_remap_job_id__\": null}",
+                        "tool_uuid": null,
+                        "tool_version": "1.1.0",
+                        "type": "tool",
+                        "uuid": "10bb239a-98ea-42df-b481-fcb1311da260",
+                        "when": null,
+                        "workflow_outputs": []
+                    },
```
wm75: You should be able to replace this renaming dance with a single APPLY_RULES step.

workflows/microbiome/samestr-workflow/samestrgal.ga
```diff
@@ -0,0 +1,1299 @@
...
+                            "name": "samestr",
+                            "owner": "iuc",
+                            "tool_shed": "toolshed.g2.bx.psu.edu"
+                        },
+                        "tool_state": "{\"clade_settings\": {\"clade_min_samples\": \"2\", \"marker_selection\": {\"retain_or_drop\": \"\", \"__current_case__\": 0}, \"marker_trunc_len\": \"20\"}, \"database\": {\"__class__\": \"ConnectedValue\"}, \"global_position_filtering\": {\"global_pos_min_n_vcov\": \"2\", \"global_pos_min_f_vcov\": \"0.0\"}, \"input_names\": {\"__class__\": \"ConnectedValue\"}, \"input_snvs\": {\"__class__\": \"ConnectedValue\"}, \"output_settings\": {\"keep_poly\": false, \"keep_mono\": false, \"delete_pos\": false}, \"sample_filtering\": {\"samples_select\": {\"__class__\": \"RuntimeValue\"}, \"samples_min_n_hcov\": \"5000\", \"samples_min_f_hcov\": \"0.0\", \"samples_min_m_vcov\": \"0.0\"}, \"sample_position_filtering\": {\"sample_pos_min_n_vcov\": \"1\", \"sample_pos_min_sd_vcov\": \"3.0\"}, \"sample_variant_filtering\": {\"sample_var_min_n_vcov\": \"2\", \"sample_var_min_f_vcov\": \"0.1\"}, \"__page__\": 0, \"__rerun_remap_job_id__\": null}",
```
wm75: Currently, you're hard-coding all the filter parameters (along with the alignment parameters in SameStr Convert and the comparison settings in SameStr Compare.
Is that really useful or does the user need control over these settings?

workflows/microbiome/samestr-workflow/samestrgal.ga
```diff
@@ -0,0 +1,1299 @@
...
+                        ],
+                        "size": [
+                            244,
+                            504
+                        ],
+                        "type": "frame"
+                    }
+                ],
+                "format-version": "0.1",
+                "name": "Metaphlan Subworkflow",
```
wm75: You'd want to add "creator" and "license" also to subworkflows so that this information doesn't get lost when the subworkflow is used in other contexts.

wm75: I'm also not sure whether this particular sub-workflow is structured correctly.
I guess you'd want to re-use its SameStr part in case the user wants both a metaphlan *and* a motus-based analysis? But then it doesn't make sense to have the metaphlan step inside the sub-workflow.

workflows/microbiome/samestr-workflow/samestrgal.ga
```diff
@@ -0,0 +1,1299 @@
...
+            "class": "Person",
+            "familyName": "Morera Martínez",
+            "name": "Xenia",
+            "identifier": "https://orcid.org/0009-0000-2380-8385"
+        }
+    ],
+    "format-version": "0.1",
+    "license": "MIT",
+    "name": "SamestrGal",
+    "release": "0.1",
```
wm75, line 112: The workflow should also have a bundled "readme", which can be a simple copy of the README file.

workflows/microbiome/samestr-workflow/samestrgal.ga
xens25: Originally both MetaPhlAn and mOTUs subworkflows were included in the main workflow so users could select which tool to run. I guess without both the MetaPhlAn subworkflow alone doesn't make much sense. I initially tried to use the SameStr steps as the subworkflow to reuse the steps but the problem I have with this, is that since only one profiler is supposed to run, not both at the same time, I implemented a boolean input + a conditional skip for both so the user could select the preferred one. Since SameStr has required inputs it gives me and error saying cannot connect optional outputs to non-optional inputs. Is this what you meant, or what approach are you suggesting instead?

workflows/microbiome/samestr-workflow/samestrgal.ga
xens25: Yes, it would be good if the user could modify them, especially for SameStr Convert and SameStr Filter. Right now the hard coded values match the ones that the original SameStr paper used. The only "issue" I see with making them workflow input parameters is that SameStr Filter alone has about 15 modifiable parameters, and I'm not sure if that would be too confusing for users.  Do you have any suggestions or alternatives @wm75 ?

bebatut:
Looks like a really interesting and needed workflow. Thanks @xens25!

@mvdbeek, @wm75: Should we combine it with the metagenomic taxonomic community profiling workflow (not yet finished, because of DB issues, #1063)? I see pros and cons for it, just would like your opinions about that.

workflows/microbiome/samestr-workflow/samestrgal.ga
wm75: If you think all 15 or so parameters are worth exposing, then you probably should.
Make sure you are arranging the input param boxes in a way that they get presented logically on the workflow run form. Distance to the upper-left corner in the wf editor is what determines the order in the WF run form, and you can of course use annotations to make it clear which parameters is for what, provide the default values from the paper, etc.

workflows/microbiome/samestr-workflow/samestrgal.ga
wm75: so if you don't need to cover the case that both should run, then you don't need a sub-workflow here at all I think.
You would have the boolean + conditional skip that you're describing, and would simply add a Pick Param step that takes the metaphlan or motus output that was actually produced. The output of the pick param tool is non-optional so can be passed to SameStr without an issue.

wm75:
> Looks like a really interesting and needed workflow. Thanks @xens25!
> 
> @mvdbeek, @wm75: Should we combine it with the metagenomic taxonomic community profiling workflow (not yet finished, because of DB issues, #1063)? I see pros and cons for it, just would like your opinions about that.

That might make sense if one can get the conditional branches right for all possible use cases. I guess I'd pass that question on to @SaimMomin12

SaimMomin12:
> Looks like a really interesting and needed workflow. Thanks @xens25!
> 
> @mvdbeek, @wm75: Should we combine it with the metagenomic taxonomic community profiling workflow (not yet finished, because of DB issues, #1063)? I see pros and cons for it, just would like your opinions about that.

Hi @bebatut, the goal of bringing this workflow into Galaxy is to provide a 1:1 Galaxy-based equivalent of the original Samestr workflow for the NFDI4Microbiome consortium. As such, @xens25 as part of her thesis ported the exact Nextflow implementation into Galaxy (as we did in the past for MGnify).

Of course, once it's part of the Galaxy ecosystem, you're welcome to reuse its components in your workflow.

xens25:
@wm75 I have updated the structure quite a bit, and it currently has no subworkflows. Let me know if there are further corrections to be made. Thanks in advance.

wm75:
@xens25 yes, that's about what I had in mind.
I modified your version a bit further here: https://usegalaxy.eu/published/workflow?id=8fab1b7bba26ac48
That new version:
- removes the Run MetaPhlAn boolean
- instead uses *optional* selects of a MetaPhlAn or a mOTUs DB

This creates two connections that the WF editor warns about, but currently there is no workaround for having optional input that are also restricted by their downstream step.
The current logic then is:
- run either MetaPhlAn or mOTUs depending on which DB the user chooses
- if they select *both* types of DBs both tools are run, but SameStr will work with the MetaPhlAn output
- things will fall apart if the user selects no DB at all, but this (to do!) could be made clear in the input param annotations 

I also improved param sorting on the WF run page further and prefixed all remaining params with the step they affect (Preprocessing:, SameStr Convert) because I like how you did this for SameStr Filter and Summarize params.

See if you like these changes, then adjust the README (both file and WF internal version) and the test accordingly.

wm75:
One more thing: I kept the Unzip Collection step upstream of mOTUs unconditional. It keeps the WF simpler and this step is very cheap to run. You may want to hide the outputs though.

xens25:
Thanks for all the fixes @wm75, I have imported your modified workflow and added a couple things more to it. Let me know if any other fix is needed.

workflows/microbiome/samestr-workflow/samestrgal.ga
```diff
@@ -23,95 +23,79 @@
...
             "child_steps": [
-                36
+                21,
+                24
             ],
-            "color": "green",
+            "color": "yellow",
             "data": {
-                "title": "SameStr Output"
+                "title": "Preprocessing Clade and marker setting steps"
```
wm75, line 11: This box title doesn't make much sense when all you're running inside is KneadData. "QC and de-hosting" might be better.

workflows/microbiome/samestr-workflow/samestrgal.ga
```diff
@@ -120,14 +104,14 @@
...
-    "readme": "# SamestrGal for shared-strain detection in metagenomes\n\nThis workflow processes paired-end shotgun metagenomic sequencing reads from multiple samples. It performs the following steps:\n\n- **Preprocessing**: Quality trimming and host read removal using KneadData\n- **Taxonomic profiling and marker-based alignment**: MetaPhlAn4 or mOTUs for taxonomic classification and marker based alignment\n- **Strain detection**: SameStr tools identify strains across samples and compute pairwise strain comparisons using Maximum Variant Profile Similarity (MVS) scores.\n\n## Inputs\n\n- **Raw Reads**: A collection of paired-end FASTQ files from one or more samples to compare\n- **Run MetaPhlAn**: Boolean parameter (default: true) that enables MetaPhlAn-based strain detection\n- **Run mOTUs**: Boolean parameter (default: false) that enables mOTUs-based strain detection\n\n## Outputs\n\n - **MetaPhlAn results** (if enabled): SNV profile summary statistics, taxon counts, strain events, and co-occurrence tables\n - **mOTUs results** (if enabled): SNV profile summary statistics, taxon counts, strain events, and co-occurrence tables\n",
+    "name": "SamestrGal Workflow",
```
wm75: I'm still not happy with the name.
What about: "SameStr-based shared strain detection from paired-end metagenomic data"?
Reads much more like the names of other iwc workflows. Needs to be changed here *and* in the README file.

workflows/microbiome/samestr-workflow/samestrgal.ga
```diff
@@ -120,14 +104,14 @@
...
...sequencing reads from multiple samples. It performs the following steps:\n\n- **Preprocessing**: Quality trimming and host read removal using KneadData\n- **Taxonomic profiling and marker-based alignment**: MetaPhlAn4 or mOTUs for taxonomic classification and marker based alignment\n- **Strain detection**: SameStr tools identify strains across samples and compute pairwise strain comparisons using Maximum Variant Profile Similarity (MVS) scores.\n\n## Inputs\n\n- **Raw Reads**: A collection of paired-end FASTQ files from one or more samples to compare\n- **Select MetaPhlAn Database**: Optional. Selecting a database runs MetaPhlAn as the taxonomic profiler and marker-based aligner.\n- **Select mOTUs Database**: Optional. Selecting a database runs mOTUs as the taxonomic profiler and marker-based aligner.\n- At least one of the two databases must be selected, or the workflow will fail. If both are selected, both tools run, but SameStr uses the MetaPhlAn output.\n\n## Outputs\n\n - **SameStr results**: SNV profile summary statistics, taxon counts, strain events, and co-occurrence tables, generated from MetaPhlAn output if a MetaPhlAn database was selected, otherwise from mOTUs output.",
```
wm75: > # SamestrGal for shared-strain detection in metagenomes

# SameStr-based shared strain detection from paired-end metagenomic data

workflows/microbiome/samestr-workflow/README.md
wm75: SameStr-based shared strain detection from paired-end metagenomic data (cmp in-wf comments)

workflows/microbiome/samestr-workflow/samestrgal-tests.yml
```diff
@@ -0,0 +1,104 @@
...
+        elements:
+        - class: File
+          identifier: forward
+          location: https://zenodo.org/records/20745835/files/28A_R1.fastq.gz
+          filetype: fastqsanger.gz
+        - class: File
+          identifier: reverse
+          location: https://zenodo.org/records/20745835/files/28A_R2.fastq.gz
+          filetype: fastqsanger.gz
+    "Preprocessing: Select Host reference genome": hg19
```
wm75: Use the yeast genome `sacCer3` here. It's much smaller, i.e. test-friendlier, and exists on cvmfs.

wm75 (commented):
Nice work @xens25 !
Just a few remaining changes.

wm75:
Ok, a bit more things to do @xens25, says the AI.  Take a look also at the suppressed comments. Those are valid ones, too, I think.

SaimMomin12:
@mvdbeek @wm75 Is this workflow ready to be merged? I would like to showcase this workflow to our potential collaborators.

mvdbeek:
I mean still blocked on https://github.com/galaxyproject/idc/pull/89 but that shouldn't stop you from demoing it ?

PR #1275, opened 2026-07-10
Convert duplicate-keyed asserts blocks to list form (passing subset)

jmchilton:

## Summary

YAML silently collapses duplicate mapping keys, so `asserts:` blocks that repeat an assertion type (`has_text:`, `has_line:`, `has_text_matching:`) only ran the **last** assertion per type — every earlier assertion was silently dropped by the loader and never executed. This converts the affected workflow-tests files to list form so every assertion runs:

```yaml
# Before — second has_text: overwrites first, "gene" assertion never ran
asserts:
  has_text:
    text: "gene"
  has_text:
    text: "BUSTED"

# After — both run
asserts:
  - that: has_text
    text: "gene"
  - that: has_text
    text: "BUSTED"
```

Successor to closed #1217, reduced to the files whose tests are **genuinely green** with this change. Files whose tests fail/error are split into their own PRs.

## Files (6)

- `bacterial_quality_and_contamination_control_post_assembly-tests.yml`
- `quality_and_contamination_control_raw_reads-tests.yml`
- `host-or-contamination-removal-on-long-reads-tests.yml`
- `short-read-quality-control-and-trimming-tests.yml`
- `pe-artic-variation-tests.yml`
- `genotype-variant-calling-wgs-pe-test.yml`

## Split out

- **#1276** — bacterial-genome-assembly (a resurrected assertion was wrong; fixed there)
- **#1277** — se-wgs-variation: removed from this set because it currently errors on a pre-existing `bcftools concat` SIGSEGV (unrelated to the assertion conversion). IWC only tests changed workflows, so touching that file merely surfaced the crash.

## How this was found

Galaxy's new `gxwf validate-tests` validator swept IWC workflow-tests against the Pydantic `Tests` model; `ordered_load` raises on duplicate mapping keys (per YAML 1.2), surfacing these authoring bugs.

FOR CONTRIBUTOR:

### DIFF
--- a/workflows/bacterial_genomics/bacterial-quality-and-contamination-control-post-assembly/bacterial_quality_and_contamination_control_post_assembly-tests.yml
+++ b/workflows/bacterial_genomics/bacterial-quality-and-contamination-control-post-assembly/bacterial_quality_and_contamination_control_post_assembly-tests.yml
@@ -28,10 +28,10 @@
           n: 6
     Kraken2 sequence assignation:
       asserts:
-        has_text:
+        - that: has_text
           text: "contig00001"
-        has_text:
-          text: "contig00359"          
+        - that: has_text
+          text: "contig00359"
     Kraken2 tabular report:
       asserts:
         has_text:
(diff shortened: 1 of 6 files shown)


mvdbeek (approved):

mvdbeek:

https://github.com/galaxyproject/iwc/actions/runs/29231569359

PR #1277, opened 2026-07-10
se-wgs-variation: convert asserts to list form (blocked by bcftools concat SIGSEGV)

jmchilton:

## Summary

Convert the `se-wgs-variation` `asserts:` blocks from duplicate-keyed dict form to list form so every assertion runs (same fix as the passing-subset PR). Split out on its own because the test currently **errors for a reason unrelated to this change**, and I didn't want it dragging the green set red.

## Why this is red (not our change)

IWC CI only tests **changed** workflows, so editing this test file is what caused `se-wgs-variation` to run at all — surfacing a pre-existing crash:

```
CRITICAL: command failed with return code -11:
bcftools concat -a -O z -o concat.vcf.gz 0.vcf.gz 1.vcf.gz 2.vcf.gz
```

`rc -11` = **SIGSEGV**. `bcftools concat` segfaults inside `lofreq_call`'s biocontainer while combining the (empty — "0 tests performed", "idxstats reports no reads mapped") VCF chunks. That's a Galaxy job-level crash, upstream of anything the test YAML does — the assertion conversion cannot cause it. The workflow errors before assertions are ever evaluated, so this change can't be green-validated until the crash is resolved.

## Tracking

- [ ] Determine whether the `bcftools concat` SIGSEGV is biocontainer/version-specific or a broader environment regression (it may affect other bcftools/lofreq workflows — only changed ones get tested, so it's under-reported)
- [ ] Once the crash is fixed, confirm this workflow's assertions pass and fold this in

Do **not** merge while red — this PR is a holding place for the assertion fix + a paper trail on the crash.

### DIFF
--- a/workflows/sars-cov-2-variant-calling/sars-cov-2-se-illumina-wgs-variant-calling/se-wgs-variation-tests.yml
+++ b/workflows/sars-cov-2-variant-calling/sars-cov-2-se-illumina-wgs-variant-calling/se-wgs-variation-tests.yml
@@ -16,11 +16,11 @@
       element_tests:
         SRR11605118:
           asserts:
-            has_line:
+            - that: has_line
               line: "##fileformat=VCFv4.0"
-            has_line:
+            - that: has_line
               line: "#CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO"
-            has_text_matching:
+            - that: has_text_matching
               expression: "NC_045512.2\t241\t.\tC\tT\t[0-9.]*\tPASS"
-            has_text_matching:
+            - that: has_text_matching
               expression: "NC_045512.2\t16111\t.\tC\tT\t[0-9.]*\tPASS\tDP=[0-9]*;AF=0.[89][0-9]*;SB=0;DP4=[0-9]*,[0-9]*,[0-9]*,[0-9]*;EFF=SYNONYMOUS_CODING\\(LOW|SILENT|Cta/Tta|L5283|7096|ORF1ab|protein_coding|CODING|GU280_gp01|2|T\\),SYNONYMOUS_CODING\\(LOW|SILENT|Cta/Tta|L891|931|ORF1ab|protein_coding|CODING|YP_009725307.1|2|T|WARNING_TRANSCRIPT_NO_START_CODON\\)"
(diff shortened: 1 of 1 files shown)


jmchilton:
> _Posted by Claude (AI assistant) on behalf of @jmchilton — not authored by them personally._

## Root cause found + upstream fix filed

The red here is **not** the assert conversion — it's a pre-existing crash this workflow surfaces (IWC only tests changed workflows). Tracked it all the way down:

**Chain:** `lofreq call-parallel` splits the reference into regions, calls per region, and merges the per-region VCFs with `bcftools concat -a` (`lofreq2_call_pparallel.py`, `concat_vcf_files`). The SE test data leaves some regions with no reads / no called variants → **header-only VCFs** → `bcftools concat -a` **segfaults (SIGSEGV, rc -11)**.

**Why:** it's a bcftools regression — [samtools/bcftools#2286](https://github.com/samtools/bcftools/issues/2286), *"bcftools concat segfault if VCF files contain only headers."* Present in **1.21**, fixed in **1.22**. The pinned `lofreq_call/2.1.5` biocontainer bundles exactly **bcftools 1.21**, and `lofreq` calls `bcftools` from `PATH`, so it inherits the crash.

Verified in the shipped container (`lofreq:2.1.5--py310h4966b78_15`), same inputs/command:

| bcftools | `concat -a` on header-only VCFs |
|---|---|
| 1.20 | ✅ exit 0 |
| **1.21** (shipped) | ❌ **SIGSEGV (exit 139)** |
| 1.22 | ✅ exit 0 |

**Fix filed upstream:** bioconda-recipes PR **[bioconda/bioconda-recipes#67152](https://github.com/bioconda/bioconda-recipes/pull/67152)** — pins `bcftools >=1.22` in the `lofreq` recipe and bumps the build number. No lofreq code change needed.

Once that merges and the lofreq package rebuilds with bcftools ≥1.22, this workflow's `bcftools concat` should stop crashing and this PR's list-form conversion can be validated green. Keeping this PR parked until then.

mvdbeek (approved):
Can you take it out of draft @jmchilton ?

mvdbeek:
I'm gonna assume it's ready, we can always do another follow up. Thank you so much @jmchilton!

mvdbeek:

https://github.com/galaxyproject/iwc/actions/runs/29726466781

PR #1279, opened 2026-07-11
capheine-core-and-compare: convert asserts to list form + relocate misplaced MEME assert

jmchilton:

## Summary

Two commits, now **green** in CI:

1. **Convert `asserts:` to list form** — the repeated `has_text:` / `has_n_columns:` keys silently collapsed to the **last** one under YAML, so only one assertion per output was validated.
2. **Relocate the resurfaced `MEME` assertion** — see below.

## The MEME fix

Converting to list form revived a `has_text: "MEME"` assertion on `combined_summary` that never actually ran before. It failed because **MEME genuinely isn't in that output** — and it never should be:

- `combined_summary` (produced by DRHIP) is **gene-level**: `gene`, `BUSTED`, `RELAX`, `pval`, `omega`. The [drhip tool's own tests](https://github.com/galaxyproject/tools-iuc/blob/main/tools/drhip/drhip.xml) assert exactly those and **never** MEME.
- MEME is a **site-level** selection test; DRHIP puts its results in `combined_sites`, in a column the tool spells **`meme_marker`** (lowercase). A case-sensitive `has_text: "MEME"` could never match.

So the assertion was dead code that the duplicate-key collapse happened to hide. Fix: drop it from `combined_summary` and add `combined_sites: has_text: meme_marker` (matching DRHIP's own convention) so the check actually verifies MEME contributed.

✅ CI is green.

### DIFF
--- a/workflows/comparative_genomics/hyphy/capheine-core-and-compare-tests.yml
+++ b/workflows/comparative_genomics/hyphy/capheine-core-and-compare-tests.yml
@@ -35,17 +35,17 @@
   outputs:
     combined_summary:
       asserts:
-        has_text:
+        - that: has_text
           text: "gene"
-        has_text:
+        - that: has_text
           text: "BUSTED"
-        has_text:
+        - that: has_text
           text: "MEME"
-        has_text:
+        - that: has_text
           text: "pval"
-        has_text:
+        - that: has_text
           text: "omega"
-        has_n_columns:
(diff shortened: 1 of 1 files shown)


d-callan (approved):

mvdbeek:

https://github.com/galaxyproject/iwc/actions/runs/29276465561

Issue #1294, opened 2026-07-14
rnaseq-sr 'featurecounts not installed' is a CI harness bug: shed data-table conf lost across Galaxy restarts


jmchilton:
> [!NOTE]
> Posted by Claude (AI assistant) on behalf of @jmchilton — not authored by them personally.

## Summary

`rnaseq-sr` has been failing in every weekly run with what looks like a Toolshed problem:

```
Workflow was not invoked; the following required tools are not installed:
toolshed.g2.bx.psu.edu/repos/iuc/featurecounts/featurecounts/2.1.1+galaxy1
```

It is **not** a Toolshed problem, and it is not a problem with the workflow or the tool. `featurecounts 2.1.1+galaxy1` is present, downloadable and valid in the Toolshed, and loads fine on usegalaxy.org.

It is a bug in our CI harness: **the second and later consumer of a shed-installed-data-table tool within a chunk cannot load that tool.** Today that set has exactly one member, `rnaseq-sr`.

## What actually happens

Galaxy fails to *parse* the tool:

```
ValueError: invalid literal for int() with base 10: 'dbkey'
  galaxy/tools/parameters/dynamic_options.py::column_spec_to_index
```

featurecounts.xml declares a filter on a **named** column:

```xml
<options from_data_table="featurecounts_anno">
    <filter type="data_meta" key="dbkey" ref="alignment" column="dbkey"/>
```

That name only resolves if the `featurecounts_anno` / `gene_sets` data tables are registered — they are what define `<columns>value, dbkey</columns>`. Those tables ship in the tool's own `tool_data_table_conf.xml.sample` and are registered by the **shed install**. If they are not registered, `column_spec_to_index` falls through to `int("dbkey")` and throws, the tool never loads, and Galaxy reports it as "not installed" at invocation.

## Why it only hits the second workflow

[`planemo_ci_actions.sh:7`](https://github.com/galaxyproject/planemo-ci-action/blob/main/planemo_ci_actions.sh#L7) persists the shed **tool** conf across Galaxy restarts:

```
--shed_tool_conf /tmp/shed_tool_conf.xml --shed_tool_path /tmp/shed_dir
```

[...]

PR #1292, opened 2026-07-14
tests: declare class:Collection / collection_type (variant calling, omics, imaging)

jmchilton:
Schema regularization of workflow test files for variant calling, transcriptomics, scRNA-seq, imaging, proteomics, and metabolomics. 13 workflow directories, all green in #1286's CI.

## What this is

Split out of #1286, which bundled 55 workflow directories into one PR and could not complete CI (the PR test job caps at 4 chunks, so ~14 Galaxy test runs landed in each and chunk 0 was cancelled at GitHub's 6h job limit).

The change is purely mechanical schema regularization of workflow test files, in two patterns:

- `class: Collection` added to outputs that already carried `element_tests:` / `elements:` / collection-only `attributes:` — a declarative annotation of things already known to be collections.
- `type:` renamed to `collection_type:` on nested collection elements in job inputs.

No `.ga` workflow files and no expected test data are touched.

## Why this one is expected green

Every workflow in this PR **passed** its planemo test run on the #1286 branch (CI run [29268785765](https://github.com/galaxyproject/iwc/actions/runs/29268785765)), and all directories linted clean. This PR carries only those workflows, in a small enough set that CI can actually finish.

Workflows in #1286 that failed did so for pre-existing reasons unrelated to this schema change (a broken `compleasm` tool, tool revisions missing from the Toolshed, stale expected VCFs, flaky asserts). Those are held back in separate branches behind their underlying fixes.

## Workflows

Plot-Nx-Size, tissue-microarray-analysis, mfassignr, openms-metaprosip, sars-cov-2-pe-illumina-artic-ivar-analysis, sars-cov-2-pe-illumina-artic-variant-calling, fastq-to-matrix-10x, velocyto, rnaseq-pe, haploid-variant-calling-wgs-pe, ploidy-aware-genotype-calling, variation-reporting, pox-virus-amplicon

PR #1295, opened 2026-07-14
tests: declare class:Collection / collection_type (rnaseq-sr)

jmchilton:
Final piece of the #1286 split. Same mechanical change as #1290/#1291/#1292: `class: Collection` on outputs that already carried `element_tests:`, and `type:` -> `collection_type:` on nested collection elements. No `.ga` files or expected data touched.

## Why this is standalone

`rnaseq-sr` was held back from #1292 because it errors in CI with:

```
Workflow was not invoked; the following required tools are not installed:
.../featurecounts/featurecounts/2.1.1+galaxy1
```

That is **not** a tool or workflow problem — it is the CI harness bug filed as #1294. The shed-installed data-table config does not survive Galaxy restarts, so the *second and later* consumer of featurecounts within a chunk cannot load it. `rnaseq-pe` runs first, installs featurecounts, and passes; `rnaseq-sr` inherits it and fails.

As the **only** featurecounts consumer in this PR's CI run, `rnaseq-sr` installs the tool fresh and registers the data tables — so it should pass here. (`rnaseq-pe` went out separately in #1292.)

This PR therefore doubles as the empirical check on #1294's diagnosis: if it goes green, the harness bug is confirmed and `rnaseq-sr` itself is fine.


mvdbeek (approved):

mvdbeek:

https://github.com/galaxyproject/iwc/actions/runs/29392418499

PR #1296, opened 2026-07-14
tests: declare class:Collection / collection_type (baredSC)

jmchilton:
Part of the #1286 split. Same mechanical change as #1290/#1291/#1292: `class: Collection` on outputs that already carried `element_tests:`. No `.ga` files or expected data touched, and **no assertions modified**.

## Why this is standalone

`baredSC-1d-logNorm` failed once in #1286 on a PNG file-size assert (`108407 +- 10000`, saw `119622`). It **passes on `main`** in the weeklies of [2026-07-06](https://github.com/galaxyproject/iwc/actions/runs/28760707797) and [2026-07-13](https://github.com/galaxyproject/iwc/actions/runs/29215707224), so that failure was not caused by this schema change.

baredSC is an MCMC sampler, so plot output varies run to run and the `+- 10000` delta is tight enough to catch it occasionally. Sending this on its own so a flake here cannot drag an otherwise-green batch red.

Worth a separate look: widening that delta would stop this recurring. Deliberately not done here, since it means loosening an assertion.

### DIFF
--- a/workflows/scRNAseq/baredsc/baredSC-1d-logNorm-tests.yml
+++ b/workflows/scRNAseq/baredsc/baredSC-1d-logNorm-tests.yml
@@ -12,6 +12,7 @@
     Maximum number of Gaussians to study: '4'
   outputs:
     baredsc_numpy:
+      class: Collection
       element_tests:
         split_file_000000.tabular:
           asserts:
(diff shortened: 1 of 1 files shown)


lldelisle (approved):
It is fine for me to loose the assertion also.

mvdbeek:

https://github.com/galaxyproject/iwc/actions/runs/29416791724

PR #1370, opened 2026-09-15
Update fairymags wfs

paulzierep:
Do we need to bump the versions for those workflows? only metadata changed.
@bebatut can you have a look if ok with you ?


SantaMcCloud:
I think you also have to check the version if there are any update to them since this are the errors: 

```
  .. ERROR: The tool toolshed.g2.bx.psu.edu/repos/iuc/map_param_value/map_param_value/0.2.0 references changeset_revision 5ac8a4bf7a8d, which is not an installable revision of the repository.
  .. ERROR: The tool toolshed.g2.bx.psu.edu/repos/iuc/filter_tabular/filter_tabular/3.3.1 references changeset_revision 90f657745fea, which is not an installable revision of the repository.
```

Not sure if you have to bump the version for it but normaly the lint also should error because of this.

paulzierep:
@bebatut can you review ?

